PR #12, opened 2020-10-19
Fix clean argument

### LINKED ISSUE #10: paru --clean doesn't work
`paru -c` works but `paru --clean` doesn't.

The response I get is: `error: invalid option '--clean'` which is a pacman response.

I tried looking in the code and stepping through the code but I have a hard time following what's going on. Probably because I'm still quite new to Rust.

If you want I can try to fix it and create a PR if you can give me some hints.

starquake:
closes #10


starquake:
closes #10

Morganamilo:
Looks perfect to me. Thanks.

Issue #8, opened 2020-10-19
-Gp should url encode package names


Morganamilo:
```
morganamilo@Octavia master ~/aur/paru % paru -Gp libc++
error: libc++: https://aur.archlinux.org/cgit/aur.git/plain/PKGBUILD?h=libc++: 404 Not Found
```

The url should be: `https://aur.archlinux.org/cgit/aur.git/plain/PKGBUILD?h=libc%2b%2b`

PR #15, opened 2020-10-21
Fixed URL encoding issues in getpkgbuilds option

haxie-dev:
Fix for issue #8


Morganamilo:
Although this works, I've never heard or urlencoding before. I could suggest using the popular percent encoding library. Not only because it's more popular. But the URL crate already depends on it. So using it wont be pulling in a new dep.

haxie-dev:
I see. Understood, I'll close then

Issue #19, opened 2020-10-24
Allow customising devel prefixes


Morganamilo:

ashhaxie:
Can you add more information on this? @Morganamilo

Morganamilo:
Paru uses a predefined list of suffixes to decide is a package is a devel package. This is used for checking if we need to try a pkgver bump when using --needed. It could be useful to make this a configurable option.

PR #26, opened 2020-10-26
Config opts for Devel Suffixes

ashhaxie:
added the config option and commandline flag for Devel Suffixes, resolving #19


micwoj92:
Am I missing something or -hg isn't there?

PR #31, opened 2020-10-28
Order results of -Ss using BottomUp/TopDown

mrkajetanp:
Currently paru -Ss shows results top-down no matter whether the BottomUp option is set or not.
This patch makes it use the same approach the interactive mode already is using to ensure consistency across different paru commands.


Morganamilo:
Didn't expect this so fast. Looks perfect, thanks.

mrkajetanp:
You're welcome, thanks for a great tool!

PR #33, opened 2020-10-28
Switch from anyhow to eyre

aspenluxxxy:
With [eyre](https://crates.io/crates/eyre) and [color-eyre](https://crates.io/crates/color-eyre), we can get better-looking error messages, while retaining the functionality of anyhow.

Example of color-eyre: 
image

(I use Debian in WSL so I can't build this and purposefully error to take a screenshot... as it links against libalpm)


Morganamilo:
I'm quite happy with the error printing as it is now. The goal it to be pacman like. So `error:` followed by `message 1: message 2` Like so:

```
morganamilo@Octavia #34 ~git/paru % paru -Ssa a     
error: aur search failed: Query arg too small.
```

If the error messages look anything like in the picture I think it goes against the goals of this project.

aspenluxxxy:
Thanks for the explanation!

PR #36, opened 2020-10-28
Improve wording and spelling in documentation.

Janfel:


Morganamilo:
Thanks, I like most of the changes, however I do prefer pkgbuilds being lowercase. Sure the PKGBUILD file is in all caps but I like to think I'm talking about the concept of a pkgbuild.

Then again https://wiki.archlinux.org/index.php/PKGBUILD uses it in all caps everywhere so I don't know.

Morganamilo:
Even though I think it's kinda ugly. It is capitalised everywhere so I guess this is more correct.

In that case the error messages in download.rs need caps too.

Morganamilo:
looks all good to me now.

PR #37, opened 2020-10-28
Fix wrong local version for repo packages when using --upgrademenu

davidnemec:


Morganamilo:
Thanks for finding. I have a feeling though this may crash on `replaces` seems as there will be no local version.

Morganamilo:
Oh actually looks libalpm answers no to replaces by default so paru doesn't print them. Gotta fix that first.

davidnemec:
Oh okay just let me know if there will be anything I need to do. I'm not that familiar with Rust and I was lazy to open an issue instead. I thought this would be the easy one :smile:

Morganamilo:
I've decided to just print replaces instead of making them part of the menu. So this should be safe.

PR #40, opened 2020-10-29
Man page improvements

owendaprile:
I went through the man pages after noticing that paru.conf(5) said the arguments for a few options (e.g. Redownload, Rebuild) didn't work when entered the way they appear in the man page. I found some typos, capitalization issues and formatting errors and fixed those (hopefully I caught them all). I made quite a few other changes that I think better explain some of the options too. I also tried to bring consistency to a lot of the terms (like AUR, PKGBUILD, pacman, paru). I also changed all the lines that ended in a period because man adds two spaces when that's entered.

Two things I noticed that should stay updated are the headers (at line 2) that should be kept up to date with the release version and the last time the man page was changed. Also, on line 467 of paru.8, there's a mention of go-alpm which I'm guessing is left over from the yay docs. I'm not sure what that should be changed to, if there's an equivalent Rust crate or it was integrated into paru.

Anyway if I got anything wrong or if there's any more I can do, just let me know.

### DIFF
--- a/man/paru.8
+++ b/man/paru.8
@@ -364,7 +364,7 @@ another package, install all the packages in the install queue.
 Always install AUR packages immediately after building them.
 
 .TP
-.B \-\-rebuild <yes|no|all>
+.B \-\-rebuild [yes|no|all]
 Always build target packages even when a copy is available in cache. If all is
 selected then all packages are rebuilt, not only targets. Defaults to no.
 
--- a/man/paru.8
+++ b/man/paru.8
@@ -1,9 +1,9 @@
 '\ t
-.TH "PARU" "8" "2019\-10\-21" "paru v0.0.1" "Paru  Manual"
+.TH "PARU" "8" "2020\-10\-28" "paru v1.0.0" "Paru Manual"
 .nh
 .ad l
 .SH NAME
-PARU \- AUR Helper
+PARU \- An AUR helper and pacman wrapper
 
 .SH SYNOPSIS
 \fIparu\fR <operation> [options] [targets]
@@ -13,7 +13,8 @@ PARU \- AUR Helper
 \fIparu\fR
 
 .SH DESCRIPTION
-Paru is an AUR helper and pacman wrapper.
+Paru is an AUR helper written in Rust and based on the design of yay. It aims
+to be your standard pacman wrapping AUR helper with minimal interaction.
 
 Paru is a tool to easily build and install packages from the AUR, along with
 their dependencies. Paru also expands many of pacman's options making them
@@ -31,7 +32,7 @@ Printing related options.
 
 .TP
 .B \-G, \-\-getpkgbuild
-Downloads PKGBUILD from ABS or AUR. The ABS can only be used for Arch Linux
+Downloads PKGBUILDs from ABS or AUR. The ABS can only be used for Arch Linux
 repositories
 
 .RE
@@ -48,10 +49,9 @@ Paru will also remove cached data about devel packages.
 
 .TP
 .B \-Sc
-Paru will also clean cached AUR package and any untracked Files in the
+Paru will also clean cached AUR packages and any untracked files in the
 cache. Cleaning untracked files will wipe any downloaded sources or
-built packages but will keep already downloaded vcs sources.
-
+built packages but will keep already downloaded VCS sources.
 
 .TP
 .B \-S, \-Si, \-Sl, \-Ss, \-Su, \-Qu
@@ -66,13 +66,13 @@ terms and prompts the user on which packages to install.
 
 .TP
 .B \-\-gendb
-Generate development package database. Tracks the latest commit for each
-development package, when there is a new commit paru will know to update. This
-is done per package whenever a package is synced. This option should only be
-used when migrating to paru from another AUR helper.
+Generate the development package database. This tracks the latest commit for
+each development package, so when there is a new commit paru will know to
+update. This is done per package whenever a package is synced. This option
+should only be used when migrating to paru from another AUR helper.
 
-This also causes paru to assume all current development packages are up to date.
-Updates will then be detected on the next commit.
+This also causes paru to assume all current development packages are up to
+date. Updates will then be detected on the next commit.
 
 .TP
 .B \-c, \-\-clean
@@ -89,26 +89,24 @@ and is not intended to be used directly by the user.
 
 .TP
 .B \-w, \-\-news
-Print new news from the Archlinux homepage. News is considered new if it is
+Print new news from the Arch Linux homepage. News is considered new if it is
 newer than the build date of all native packages. Pass this twice to show all
 available news.
 
 .SH GETPKGBUILD OPTIONS (APPLY TO \-G AND \-\-GETPKGBUILD)
 .TP
-
-.TP
 .B \-p, \-\-print
-Prints the pkgbuild to the terminal instead of downloading it.
+Prints the PKGBUILD to the terminal instead of downloading it.
 
 .SH NEW OPTIONS
 .TP
-.B    \-\-repo
-Assume all targets are from the repositories. Additionally Actions such as
+.B \-\-repo
+Assume all targets are from the repositories. Additionally, actions such as
 sysupgrade will only act on repository packages.
 
 .TP
 .B \-a, \-\-aur
-Assume all targets are from the AUR. Additionally Actions such as
+Assume all targets are from the AUR. Additionally, actions such as
 sysupgrade will only act on AUR packages.
 
 Note that dependency resolving will still act normally and include repository
@@ -121,7 +119,7 @@ to use https://aur.tuna.tsinghua.edu.cn/.
 
 .TP
 .B \-\-builddir <dir>
-Directory to use for Building AUR Packages. This directory is also used as
+Directory to use for building AUR Packages. This directory is also used as
 the AUR cache when deciding if paru should skip builds.
 
 .TP
@@ -144,7 +142,7 @@ The command to use for \fBgit\fR calls. This can be a command in
 Passes arguments to git. These flags get passed to every instance where
 git is called by paru. Arguments are split on whitespace before being
 passed to git. Multiple arguments may be passed by supplying a space
-separated list that is quoted by the shell
+separated list that is quoted by the shell.
 
 .TP
 .B \-\-gpg <command>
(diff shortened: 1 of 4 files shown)


man/paru.conf.5
```diff
@@ -1,53 +1,55 @@
...
 .SH DESCRIPTION
-Paru's config file. Based on the format used by \fBpacman.conf (5)\fR.
+Paru's config file. Based on the format used by 
+.BR pacman.conf (5)
 
-Paru will attempt to read the config file from the user's home directory. If that does not exist
-paru will instead attempt to read the file from the system config directory.
+Paru will attempt to read the config file from $XDG_CONFIG_HOME/paru/paru.conf. If
+$XDG_CONFIG_HOME is unset, it will attempt to read the file from 
+$HOME/.config/paru/paru.conf. If neither of those exist, it will read from the
```
Morganamilo, line 17: You're free to edit /etc/paru.conf, so I wouldn't say "it will read from the defaults at /etc/paru.conf".

man/paru.conf.5
```diff
@@ -1,53 +1,55 @@
...
 .ad l
 .SH NAME
-PARU \- AUR Helper
+paru.conf \- paru configuration file
 
 .SH SYNOPSIS
-/etc/paru.conf
-.sp
-~/.config/paru/paru.conf
+$XDG_CONFIG_HOME/paru/paru.conf, $HOME/.config/paru/paru.conf, /etc/paru.conf
```
Morganamilo, line 9: This should be reversed. Just because that's how makepkg.conf lists it.

man/paru.conf.5
Morganamilo, line 17: You're free to edit /etc/paru.conf, so I wouldn't say "it will read from the defaults at /etc/paru.conf".

man/paru.conf.5
```diff
@@ -89,49 +91,48 @@ PKGBUILD.
...
 .TP
 .B UseAsk
-Use pacman's --ask flag to automatically confirm package conflicts. paru lists
+Use pacman's --ask flag to automatically confirm package conflicts. Paru lists
 conflicts ahead of time. It is possible that paru does not detect
 a conflict, causing a package to be removed without the user's confirmation.
 However, this is very unlikely.
 
 .TP
-.B Redownload [= All]
```
Morganamilo, line 114: Again `[...]` means optional. So `Redownload` on it's own means yes `Redownload = All` for all.

Looking at the code, `Redownload = No` works cause it uses the same validator as --redownload. But using it doesn't make sense and I may remove it. so lets not advertise it.

man/paru.conf.5
```diff
@@ -145,27 +146,31 @@ to use https://aur.tuna.tsinghua.edu.cn/.
...
 .TP
-.B CompletionInterval = N
+.B CompletionInterval = Number
 Time in days to refresh the completion cache. Setting this to 0 will cause
 the cache to be refreshed every time, while setting this to -1 will cause the
-cache to never be refreshed.
+cache to never be refreshed. Defaults to 7.
 
 .TP
 .B RequestSplit = Number
```
Morganamilo, line 169: Again, this option was removed. Looks like the option is still there in the code, but that will be removed.

man/paru.8
```diff
@@ -234,19 +233,21 @@ it is up to the user what upgrades they skip.
 Do not show the upgrade menu.
 
 .TP
-.B \-\-removemake [yes|no|ask]
```
Morganamilo, line 237: `<...>` Means required, while `[...]` means optional. `--removemake` defaults to yes in the sense that `--removemake` = `--removemake=yes`, to get ask you do `--removemake=ask.

Morganamilo (changes requested):
Changes are good apart from where commented. Also there's been a lot of people with similar ideas so you have some conflicts.

Morganamilo:
Oh forgot to mention. The equivalent go go--alpm is [alpm.rs](https://github.com/Morganamilo/alpm.rs)

man/paru.8
```diff
@@ -198,31 +197,31 @@ separated list that is quoted by the shell.
...
 .TP
 .B \-\-upgrademenu
-Show a detailed list of updates in a similar format to VerbosePkgLists.
-Upgrades can also be skipped using numbers, number ranges or repo names.
-Additionally ^ can be used to invert the selection.
+Show a detailed list of updates in a similar format to 
+VerbosePkgLists. Upgrades can also be skipped using numbers, number ranges or
+repo names. Additionally ^ can be used to invert the selection.
 
 Additionally an argument can be passed to this flag and the menu will use
```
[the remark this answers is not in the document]
owendaprile: The only part of this that was removed was the ^ inversion operator right?

man/paru.8
```diff
@@ -198,31 +197,31 @@ separated list that is quoted by the shell.
 .B \-\-completioninterval <days>
 Time in days to refresh the completion cache. Setting this to 0 will cause
 the cache to be refreshed every time, while setting this to -1 will cause the
-cache to never be refreshed.
+cache to never be refreshed. Defaults to 7.
 
 .TP
 .B \-\-requestsplitn <number>
```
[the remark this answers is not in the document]
owendaprile: Do you mean the whole --completioninterval option? Because it still seems to exist in completion.rs at update_aur_cache().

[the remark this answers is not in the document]
owendaprile: Just saw the other comment saying this will be removed. Should I just take it out now?

owendaprile:
Quick question reguarding the --redownload option. The current man pages say it accepts yes, no, or all, but it really only accepts yes or no. I'm guessing this is a bug since all makes sense as an option. Also, yes doesn't seem to work. I tried with an install and an upgrade but even when set to yes it still just says pkgbuilds up to date.

man/paru.8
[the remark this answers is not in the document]
Morganamilo: The ability to pass an argument to `--upgrademenu`. There's no real point to it seems as you can just use `--ignore`.

man/paru.8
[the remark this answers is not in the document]
Morganamilo: I was talking about `--requestsplitn`. It seems github shows context above the line I comment on.

Morganamilo:
> Quick question reguarding the --redownload option. The current man pages say it accepts yes, no, or all, but it really only accepts yes or no. I'm guessing this is a bug since all makes sense as an option. Also, yes doesn't seem to work. I tried with an install and an upgrade but even when set to yes it still just says pkgbuilds up to date.

Yep, Is a bug.

man/paru.8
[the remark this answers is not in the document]
owendaprile: Ohh. I read this as it telling you what options are available when using the menu, so I think it makes sense to keep it, especially because people might not realize you can use ranges or repo names.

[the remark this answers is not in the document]
Morganamilo: Yes Keep it, The line I want removed it `Additionally an argument can be passed to this flag and the menu will use`.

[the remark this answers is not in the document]
owendaprile: Okay I missed that line

man/paru.8
```diff
@@ -1,9 +1,9 @@
 '\ t
-.TH "PARU" "8" "2019\-10\-21" "paru v0.0.1" "Paru  Manual"
+.TH "PARU" "8" "2020\-10\-28" "paru v1.0.0" "Paru Manual"
 .nh
 .ad l
 .SH NAME
-PARU \- AUR Helper
+PARU \- An AUR helper and pacman wrapper
```
Morganamilo, line 6: Missed this last time but I think it's better without the `An`.

owendaprile:
I think --removemake is good the way I have it now. Maybe it should specify that the default behavior is no when the option isn't specified?

man/paru.conf.5
```diff
@@ -1,55 +1,55 @@
...
-PARU \- AUR Helper
+paru.conf \- paru configuration file
 
 .SH SYNOPSIS
-$PARU_CONF
-.sp
-/etc/paru.conf
-.sp
-~/.config/paru/paru.conf
+/etc/paru.conf, $XDG_CONFIG_HOME/paru/paru.conf, $HOME/.config/paru/paru.conf
```
Morganamilo, line 9: You left you `$PARU_CONF` when you merged changes.

Morganamilo:
> I think --removemake is good the way I have it now. Maybe it should specify that the default behavior is no when the option isn't specified?

I don't really mind either way. If you can think of a nice way to phrase it then add it in.

man/paru.conf.5
owendaprile: When $PARU_CONF is set paru reads the file from $PARU_CONF/paru.conf, right?

Morganamilo: Nope, `$PARU_CONF` is the `paru.conf` file.

owendaprile: Ok. I'm going to update the description with that then.

owendaprile: When the file at $PARU_CONF doesn't exist, paru exits with an somewhat useless error `error: No such file or directory (os error 2)`. Should I mention this in the man page, or should a better error message be added?

man/paru.8
owendaprile: I say AUR like I'm spelling it out, and an is the right word since A is a vowel sound. I can still change it if you like A better though.

Morganamilo: No I mean remove the world together `UR helper and pacman wrapper`. I believe that's the kind of language man pages tend to use. See man pacman/makepkg.

man/paru.conf.5
Morganamilo: Bad error message is an issue with the program. It should be fixed there.

man/paru.8
owendaprile: Alright I brought that in line with how pacman and makepkg word theirs.

man/paru.conf.5
```diff
@@ -1,55 +1,56 @@
...
-PARU \- AUR Helper
+paru.conf \- paru configuration file
 
 .SH SYNOPSIS
-$PARU_CONF
-.sp
-/etc/paru.conf
-.sp
-~/.config/paru/paru.conf
+/etc/paru.conf, $PARU_CONF, $XDG_CONFIG_HOME/paru/paru.conf, $HOME/.config/paru/paru.conf
```
Morganamilo, line 9: `$PARU_CONF` Has the highest priority, should be first.

That's all I see wrong though.

Morganamilo:
Took a while but finally merged. Thanks.

PR #48, opened 2020-10-29
Use rustls in reqwest over openssl

aspenluxxxy:
Less native dependencies + safer in theory


Morganamilo:
Actually It's more dependencies. Paru depends on libalpm which depends on openssl. So either way we're using openssl. This just makes us also depend on rusttls.

aspenluxxxy:
Agh why do I always make PRs where I don't notice something important

PR #86, opened 2020-11-06
manpages: Remove dashes from InstallDebug

kattjevfel:
Yes, this is tiny. Yes it annoyed me enough to make this PR.


Morganamilo:
An issue is an issue.

PR #97, opened 2020-11-12
Remove whitespace when there is nothing to clean

### LINKED ISSUE #96: Trim space when clean
When running `paru --clean`, it just prints `there is nothing to do` if there are no packages to clean.
Why is it still printing a space before it?
https://github.com/Morganamilo/paru/blob/master/src/main.rs#L166

nix6839:
Close #96


nix6839:
Ref [this](https://github.com/Morganamilo/paru/issues/96#issuecomment-726041433)

PR #105, opened 2020-11-18
Removed false-flagging -debug code

ashhaxie:
Some code here would false-flag packages ending in `-debug` as debug packages, even if it is the base name of the package. this fixes that.


Geweldig:
Seems paru no longer builds after this has been merged:
```
error[E0308]: mismatched types
    --> src/install.rs:1132:41
     |
1132 |                 .filter(|pkg| !is_debug(*pkg))
     |                                         ^^^^ expected struct `alpm::Package`, found `&raur::ArcPackage````

Morganamilo:
This is what I get for trusting ashie. -_-

ashhaxie:
This is what you get for not reviewing PRs

PR #110, opened 2020-11-21
install.rs: Fix not compatible with some editors

akiirui:
Some editors such as `vscode` are exec by a bash script.

`Child.wait()` received a exit code from bash, but vscode is still running.

And the TempDir(`config.fetch.make_view()`) is delete when it goes out of scope.

Then, editor cannot found TempDir.

So, ask `"Proceed with installation?"` in it's scope.

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -258,7 +258,21 @@ async fn install_actions<'a>(
         .collect::<Vec<_>>();
 
     if let Some(ref fm) = config.fm {
-        file_manager(config, fm, &pkgs)?;
+        let has_diff = config.fetch.has_diff(&pkgs)?;
+        config.fetch.save_diffs(&has_diff)?;
+        let view = config.fetch.make_view(&pkgs, &has_diff)?;
+
+        let ret = Command::new(fm)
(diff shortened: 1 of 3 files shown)


src/install.rs
```diff
@@ -360,21 +374,6 @@ async fn install_actions<'a>(
     Ok(0)
 }
 
-fn file_manager(config: &Config, fm: &str, pkgs: &[&str]) -> Result<()> {
```
Morganamilo, line 363: Why not just make this return the view instead of ripping out the whole function?

akiirui: `view ` type is `TempDir`, need 
```

akiirui: `view` type is `TempDir`. to return view, we need

```rust
install.rs:
extern crate tempdir;
use tempdir::TempDir;

Cargo.toml:
[dependencies]
tempdir = "0.3.7"

akiirui: Is it okay now?

Morganamilo: Yep that looks good to me, thanks.

PR #114, opened 2020-11-23
Add git to dependencies

tinou98:
Git is obviously required, but it's always good to have everything working with a simple copy-paste.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Paru is an AUR helper written in Rust and based on the design of [`yay`](https:/
 ## Installation
 
 ```
-sudo pacman -S --needed base-devel
+sudo pacman -S --needed base-devel git
 git clone https://aur.archlinux.org/paru.git
 cd paru
 makepkg -si


thecakeisalie25:
Isn't git a part of base-devel? Base-devel is pretty much an implied dependency for everything aur-related iirc.

Morganamilo:
It's not part of base-devel.

Morganamilo:
Anyway, yes it's pretty obvious you need git to use git. But there's no reason not to add it right?

But then again if the user doesn't have git installed then surely they should install git as a dependency `pacman -S --asdeps git`? But now that means we now have to add it as a separate line because `base-devel` should not be installed as a dep.

And there's also the fact they may want it explicitly installed.

Yes this is all kinda trivial stuff to argue about. But it makes me want to just not deal with it and let the user choose how they want to install git.

SeerLite:
git is different from base-devel because paru doesn't really need base-devel to function by itself, while it does need git.

base-devel is a catch-all build-time dependency for PKGBUILDs. Some PKGBUILDs may not even use anything from base-devel (maybe I'm wrong but I'm pretty sure it's possible). Not all AUR packages need git either. `-git` packages mark git as a dependency expliticly.

git is used directly by paru. It's the way it gets PKGBUILDs from the AUR in the first place. It's part of the core functionality of paru, unlike base-devel.


(Also git isn't even the only way to get the PKGBUILDs, you could download snapshots directly. This isn't as convenient though.)

Morganamilo:
That's right. That's why git is listed as a dependency in the pkgbuild. This is about the install instructions.

SeerLite:
Ah, I see. Sorry! I didn't even check what this was all about :s

Anyway yeah I think it's fine how it is. Maybe adding something like "Note: Remember to have git installed" could be better though.

IlanCosman:
I vote for adding it. I know I at least copy-pasted the install instructions, which failed because I was on a fresh install of Arch without git installed. Installation should always be copy-pastable when possible imo.

PR #123, opened 2020-12-02
Allow TopDown to be set in config

Geweldig:
TopDown is currently documented in `paru.conf(5)`, but can't actually be set.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -558,6 +558,7 @@ impl Config {
         let mut ok2 = true;
 
         match key {
+            "TopDown" => self.sort_mode = "topdown".into(),
             "BottomUp" => self.sort_mode = "bottomup".into(),
             "AurOnly" => self.mode = "aur".into(),
             "RepoOnly" => self.mode = "repo".into(),


Morganamilo:
Topdown doesn't really make sense in the config file because it's the default. It's the same reason. NoDevel and such are not in the config.

The mistake would be in the man page and not the code.

Geweldig:
I would argue that it would still make sense to be able to set them in the config, as defaults can change in the future. Not being able to set it would mean it's not possible to create a config which explicitly describes the behaviour you want, resulting in the possibility of the behaviour changing in case defaults change.

Morganamilo:
The idea is to have no defaults at all. Everything should be off/false. Then we ship paru.conf with the things we want to enable.

So the default would never change, only the paru.conf will.

Geweldig:
I agree with that sentiment, but not completely in the case of `TopDown`/`BottomUp` as to me there is no clear indication which of the two would be "off" without knowing that paru defaults to `TopDown` (unlike for example the `Devel` toggle, where that would by default be off).

Geweldig:
I have opened #124 to remove the setting from the man page. I opted to do this in a different PR, so this one can be closed while keeping the discussion intact.

Morganamilo:
> I agree with that sentiment, but not completely in the case of `TopDown`/`BottomUp` as to me there is no clear indication which of the two would be "off" without knowing that paru defaults to `TopDown` (unlike for example the `Devel` toggle, where that would by default be off).

I somewhat understand, as bottomup/topdown don't exactly make sense unless you think about them a little. Basically all programs print as they go, `pacman -Ss` included, so I consider that the default. Maybe if the option was named `flipsearch` it would be more obvious what the default is.

Geweldig:
I fully agree, and also consider `TopDown` the default. However I thought it was worthwhile to bring it up.

PR #125, opened 2020-12-02
Better error message when spawning file manager fails

Tyilo:


Morganamilo:
I was about to do this myself after seeing the issue.  One thing I didn't know at the time of writing the code was you can call '.status()' which acts as spawn and wait. With that we can just have the one error message.

PR #130, opened 2020-12-06
Add example for "Packages for exclude" user input

Mephistophiles:
* Added example for "Packages for exclude" user input
  Before:

  ```
  :: Packages to exclude: 
  ```
  After:
  ```
  :: Packages to exclude: (eg: "1 2 3", "1-3")
  ```

* Added a new line after user input in function `input()`
  This is pretty visible on a smaller terminal wight
  Before:
  ```
  :: Packages to exclude: (eg: "1 2 3", "1-3") 
  ```
  After:
  ```
  :: Packages to exclude: (eg: "1 2 3", "1-3")
  :: 
  ```

### DIFF
--- a/src/search.rs
+++ b/src/search.rs
@@ -286,7 +286,7 @@ pub async fn search_install(config: &mut Config) -> Result<i32> {
         }
     }
 
-    let input = input(config, "Packages to install (eg: 1 2 3, 1-3 or ^4): ");
+    let input = input(config, "Packages to install (eg: 1 2 3, 1-3 or ^4):");
 
     if input.trim().is_empty() {
         println!(" there is nothing to do");
--- a/src/search.rs
+++ b/src/search.rs
@@ -286,7 +286,7 @@ pub async fn search_install(config: &mut Config) -> Result<i32> {
         }
     }
 
-    let input = input(config, "Packages to install (eg: 1 2 3, 1-3 or ^4):");
+    let input = input(config, "Packages to install (eg: 1 2 3, 1-3):");
 
     if input.trim().is_empty() {
         println!(" there is nothing to do");
--- a/src/upgrade.rs
+++ b/src/upgrade.rs
@@ -273,7 +273,7 @@ pub async fn get_upgrades<'a, 'b>(
         );
     }
 
-    let input = input(config, "Packages to exclude:");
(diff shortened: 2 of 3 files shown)


Morganamilo:
Your description of after doesn't seem to match what you actually did.

Also thinking about it, the inversion `^4` can probably be dropped. It was only really useful for the diff/clean menu, which paru does not have.

Mephistophiles:
> Also thinking about it, the inversion ^4 can probably be dropped. 

Done. 

> Your description of after doesn't seem to match what you actually did.

Oh, my failure. I forgot to add before/after block to the first paragraph.

src/search.rs
```diff
@@ -286,7 +286,7 @@ pub async fn search_install(config: &mut Config) -> Result<i32> {
         }
     }
 
-    let input = input(config, "Packages to install (eg: 1 2 3, 1-3 or ^4): ");
+    let input = input(config, "Packages to install (eg: 1 2 3, 1-3 or ^4):");
```
Morganamilo, line 289: The ^4 can go here too.

Mephistophiles: You mean delete the "or ^4" from search ask input?

Morganamilo: yeah, I mean't "can go" as in "can be removed"

Mephistophiles: Done. Please review it

src/upgrade.rs
```diff
@@ -273,7 +273,7 @@ pub async fn get_upgrades<'a, 'b>(
         );
     }
 
-    let input = input(config, "Packages to exclude: ");
+    let input = input(config, "Packages to exclude: (eg: \"1 2 3\", \"1-3\")");
```
Morganamilo, line 276: Here you've quoted "1 2 3" while the search install has it unquoted.

Mephistophiles: OK. Did as in the search. thanks for review!

Morganamilo: There's also differences in the `:` placement.  I don't really know what I prefer, just they should be the same.

Mephistophiles: Hmmm, the semicolon at the end looks better.

Morganamilo: Looks good now.

PR #149, opened 2021-01-03
Capitalize and reword news update print

PopeRigby:
This is a really petty change but it was irking me.


Morganamilo:
The idea of this was to copy the pacman messages:

```
resolving dependencies...
looking for conflicting packages...
there is nothing to do
```

Issue #83, opened 2020-11-06
[Request] binaries for ARM (armv6, armv7, aarch etc)


m040601:
Thanks a lot for your work on this replacement for yay.

I saw this question at,

https://old.reddit.com/r/archlinux/comments/jjn1c1/paru_v100_and_stepping_away_from_yay/

> ....But I have a question it is possible to have paru-bin also available for arm architecture ? Like aarch64 or armv7 ?...

And I already know your answer,

> ....I would like to. But I'd need some sort of Ci+vm setup and that's gonna take a bit of research to get up and running.....

Like yay-bin, paru-bin, https://aur.archlinux.org/packages/paru-bin will only be much more usefull in low speced arm platforms (raspberry pi etc) once precompiled binaries are available.

So please leave this issue open, until this is done.
In the mean time, suggestions can be made here, on how to help you achieve this with the least amount of effort.

m040601:
I assume you are aware of github actions right ?

Here's some other suggestions.

Perhaps you can also have a look at how other cli rust tools do it, for example, skim, https://github.com/lotabout/skim/releases, ripgrep, https://github.com/BurntSushi/ripgrep or fd, https://github.com/sharkdp/fd

Morganamilo:
Yes I'm aware of github actions. Thing is I like to remain agnostic to the hosting site. So travis would be my go to. Except travis doesn't have a nice way to choose arch. You can use docker with it but it looks kinda annoying.

So yeah, I'm just not sure what exactly what to go for.

Morganamilo:
I ended up using github actions in https://github.com/Morganamilo/alpm.rs so I'll probably concede and use them here eventually.

patrickelectric:
Please, that is a must thing, it takes forever to compile under a raspberry pi 4.

Morganamilo:
Isn't the pi4 aarrch64? there are binaries for that.

m040601:
> Isn't the pi4 aarrch64? there are binaries for that.

What binaries are you refering to ? I haven't seen any released on github ?

Could you update us on your plans / intentions for providing them here on github ?

> it takes forever to compile under a raspberry pi 4.

I hope you understand, why people require this. It's ok, if you do not intend of using github actions for automating and providing binaries here **for arm**. But please clarify.

Morganamilo:
> What binaries are you refering to ? I haven't seen any released on github ?

The releases have had them for quite some time.

[...]

PR #166, opened 2021-01-17
Add support GitHub Actions

Mephistophiles:
This PR introduce:

- fix clippy errors
- added GitHub Actions support:
  - Every push check:
    - run cargo-test (compile test and launch test suits)
    - run cargo-clippy (check additional lints)
    - run cargo-audit (check for rust vulnerabilities)
    - run cargo-fmt (check code format)
  - Every PR check

The jobs use an ubuntu-latest image with an archlinux docker inside.

P.S. cargo clippy --all-features failed (alpm package failed)

TODO:

- [ ] Remove Github tokens from PR
- [ ] Add tests for git features (by example https://github.com/Morganamilo/alpm.rs/blob/master/.github/workflows/alpm.yml)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,78 @@
+name: Rust
+
+on:
+  pull_request:
+  push:
+
+env:
+  CARGO_TERM_COLOR: always
+  CI: 1
+
+jobs:
+  cargo-audit:
+    name: Audit Rust vulnerabilities
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: rust-audit-check
+      uses: actions-rs/audit-check@v1.2.0
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+
+  cargo-test:
+    name: Cargo Test
+    runs-on: ubuntu-latest
+
+    container:
+      image: archlinux:base-devel
+
+    steps:
+    - name: Install Packages
+      run: pacman -Syu rust clang gcc --noconfirm --needed
+    - name: Checkout repository
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Compile
+      run: cargo test --no-run --locked
+
+    - name: Test
+      run: cargo test
+
+  cargo-fmt:
+    name: Code Format
+    runs-on: ubuntu-latest
+
+    container:
+      image: archlinux:base-devel
+
+    steps:
+    - name: Install Packages
+      run: pacman -Syu rust clang gcc --noconfirm --needed
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Check format
+      run: cargo fmt --all -- --check
+
+  cargo-clippy:
+    name: Clippy
+    runs-on: ubuntu-latest
+
+    container:
+      image: archlinux:base-devel
+
+    steps:
+    - name: Install Packages
+      run: pacman -Syu rust clang gcc --noconfirm --needed
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - uses: actions-rs/clippy-check@v1
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
--- a/src/devel.rs
+++ b/src/devel.rs
@@ -311,7 +311,7 @@ pub async fn filter_devel_updates(
 
     config.raur.cache_info(cache, &updates).await?;
     let updates = updates
-        .into_iter()
+        .iter()
         .map(|u| pkgbases.remove(u.as_str()).unwrap())
         .collect::<Vec<_>>();
 
--- a/src/devel.rs
+++ b/src/devel.rs
@@ -254,7 +254,7 @@ pub async fn possible_devel_updates(config: &Config) -> Result<Vec<String>> {
     let mut pkgbases: HashMap<&str, Vec<alpm::Package>> = HashMap::new();
 
     for pkg in db.pkgs().iter() {
-        let name = pkg.base().unwrap_or(pkg.name());
+        let name = pkg.base().unwrap_or_else(|| pkg.name());
         pkgbases.entry(name).or_default().push(pkg);
     }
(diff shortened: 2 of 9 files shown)


Morganamilo:
The code changes are probably fine. The github actions stuff though I'll need some explanation of some of the bits.

.github/workflows/rust.yml
```diff
@@ -0,0 +1,99 @@
+name: Rust
+
+on:
+  pull_request:
+  push:
+    branches:
+    - master
```
Morganamilo, line 7: Id prefer this on every branch. I don't always work on master.

.github/workflows/rust.yml
```diff
@@ -0,0 +1,99 @@
...
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+
+  cargo-test:
+    name: Cargo Test
+    runs-on: ubuntu-latest
+
+    container:
+      image: archlinux:base-devel
+      options: --privileged
```
Morganamilo, line 33: What does this do and why is it needed?

.github/workflows/rust.yml
```diff
@@ -0,0 +1,99 @@
...
+      volumes:
+        - /sys/fs/cgroup:/sys/fs/cgroup
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Install Rust toolchain
```
Morganamilo, line 43: I'd prefer to just use pacman here. Less complex and has the added benefit of also testing paru compiles with the version of rust in the repos.

.github/workflows/rust.yml
```diff
@@ -0,0 +1,99 @@
...
+    runs-on: ubuntu-latest
+
+    container:
+      image: archlinux:base-devel
+      options: --privileged
+      volumes:
+        - /sys/fs/cgroup:/sys/fs/cgroup
+
+    steps:
+    - name: Install libclang (needed by clippy)
```
Morganamilo, line 85: clippy does not need clang afaik where did you get that from?

.github/workflows/rust.yml
```diff
@@ -0,0 +1,99 @@
...
+      run: sudo pacman --noconfirm -Syy clang
```
Morganamilo, line 86: Please don't -Sy or -Syy.

.github/workflows/rust.yml
```diff
@@ -0,0 +1,99 @@
...
+
+    - uses: actions-rs/toolchain@v1
+      with:
+          toolchain: nightly
+          components: clippy
+          override: true
+    - uses: actions-rs/clippy-check@v1
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        args: #--all-features (TODO: fix all features build)
```
Morganamilo, line 99: Don't just blindly test all features. There needs to be a separate test for git and non git builds.

.github/workflows/rust.yml
Mephistophiles: Ok, I'll fix that.

.github/workflows/rust.yml
Mephistophiles: The cache has been removed from the archive images archive.

```
warning: database file for 'core' does not exist (use '-Sy' to download)
warning: database file for 'extra' does not exist (use '-Sy' to download)
warning: database file for 'community' does not exist (use '-Sy' to download)
```

Morganamilo: Then use -Syu to avoid a partial upgrade.

Mephistophiles: Hmm, After installing rust from the official repository, the clang was not required.

.github/workflows/rust.yml
Mephistophiles: OK. I'll do it a little later.

.github/workflows/rust.yml
Mephistophiles: Removed.

.github/workflows/rust.yml
Mephistophiles: Switched to rust from archlinux

.github/workflows/rust.yml
Mephistophiles: Needed for using archlinux instead of ubuntu (default linux container)

Morganamilo: Works for me without it at https://github.com/Morganamilo/alpm.rs/blob/master/.github/workflows/alpm.yml

.github/workflows/rust.yml
Morganamilo: Also forgot to mention, isn't the container running as root? the sudos shouldn't be needed right?

Mephistophiles:
@Morganamilo small question: do you need the build badge in the README.md?
example: Rust

Morganamilo:
Not really, the commit already shows the ci status.

.github/workflows/rust.yml
Mephistophiles: Oh, yes. Removed.

.github/workflows/rust.yml
Mephistophiles: oh, I apologize for not immediately thinking to do so :(

.github/workflows/rust.yml
```diff
@@ -0,0 +1,78 @@
...
+    steps:
+    - name: Install Packages
+      run: pacman -Syu rust clang gcc --noconfirm --needed
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - uses: actions-rs/clippy-check@v1
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
```
Morganamilo, line 78: Also why is the secret key needed for this?

Mephistophiles: It's only needed for forked repos. I will remove it before change to "Ready for Review" state.

Morganamilo:
By the way, if you're interested in doing CI at all, It would be nice to get #83 solved.  Yay does it with this https://github.com/Jguer/yay/blob/next/.github/workflows/multiarch-build.yml but I'm not really familiar with github actions or docker.

patrickelectric:
Hey, any update ?

PR #205, opened 2021-02-02
Made TCP keepalive timeout configurable

ghuwe:
I made the TCP keepalive timeout configurable, in case 15 seconds is not enough.


Morganamilo:
Thanks but I feel there's no need for a config option for such a thing. Especially not on a just in case.

PR #216, opened 2021-02-04
Change default of BottomUp variable

halbGefressen:
I think that most people will want to use the BottomUp option anyway. So why not change the default?


Morganamilo:
The goal of paru is to be as pacman like by default with optional extras. This is why it's disabled by default.

rahatzamancse:
Is there any conf overriding flag that I can use to make it BottomUp from the commandline? Was just thinking that I can alias paru to use with that flag instead of changing the config.

PR #217, opened 2021-02-05
Make sure the script fails when a command fails

KSXGitHub:
By default, shell script will continue to execute the next command even if the previous command exited with non-zero status.


Morganamilo:
Looks good to me.

PR #240, opened 2021-02-11
Add -P --stats

vojta7:
Implement option to show system package statistics (Total installed, Total size, orphaned, ...)

### DIFF
--- a/src/stats.rs
+++ b/src/stats.rs
@@ -1,4 +1,4 @@
-use crate::config::Config;
+use crate::config::{version, Config};
 
 use alpm::PackageReason;
 use alpm_utils::DbListExt;
@@ -71,6 +71,7 @@ pub fn stats(config: &Config) -> Result<i32> {
     let c = config.color;
     let info = collect_info(config, 10)?;
 
+    version();
     print_line_separator(config);
 
     println!(
--- a/src/stats.rs
+++ b/src/stats.rs
@@ -1,4 +1,5 @@
 use crate::config::{version, Config};
+use crate::util::repo_aur_pkgs;
 
 use alpm::PackageReason;
 use alpm_utils::DbListExt;
@@ -47,7 +48,11 @@ async fn collect_info<'a>(config: &'a Config, max_n: usize) -> Result<Info<'a>> 
         total_size += pkg.isize();
     }
 
-    let aur_info = config.raur.info(&foreign_packages).await?;
+    let aur_packages = repo_aur_pkgs(config)
+        .iter()
+        .map(|pkg| pkg.name())
+        .collect::<Vec<_>>();
+    let aur_info = config.raur.info(&aur_packages).await?;
     for pkg in aur_info.into_iter() {
         if pkg.maintainer.is_none() {
             orphaned.push(pkg.name.clone());
--- a/src/stats.rs
+++ b/src/stats.rs
@@ -3,7 +3,6 @@ use crate::download::cache_info_with_warnings;
 use crate::util::repo_aur_pkgs;
 
 use alpm::PackageReason;
-use alpm_utils::DbListExt;
 
 use std::cmp::Reverse;
 use std::collections::BinaryHeap;
@@ -22,10 +21,8 @@ struct Info<'a> {
 
 async fn collect_info<'a>(config: &'a Config, max_n: usize) -> Result<Info<'a>> {
     let db = config.alpm.localdb();
-    let sync_db = config.alpm.syncdbs();
 
     let total_packages = db.pkgs().len();
-    let mut foreign_packages = Vec::new();
     let mut explicit_packages = 0;
     let mut total_size = 0;
 
@@ -35,9 +32,6 @@ async fn collect_info<'a>(config: &'a Config, max_n: usize) -> Result<Info<'a>> 
         max_packages.push(Reverse((pkg.isize(), pkg.name())));
         if max_packages.len() > 10 {
             max_packages.pop();
-        }
-        if let Err(alpm::Error::PkgNotFound) = sync_db.pkg(pkg.name()) {
-            foreign_packages.push(pkg.name());
         }
         if pkg.reason() == PackageReason::Explicit {
             explicit_packages += 1;
--- a/src/stats.rs
+++ b/src/stats.rs
@@ -48,7 +48,7 @@ async fn collect_info<'a>(config: &'a Config, max_n: usize) -> Result<Info<'a>> 
         total_size += pkg.isize();
     }
 
-    let aur_packages = repo_aur_pkgs(config)
+    let (_, aur_packages) = repo_aur_pkgs(config)
         .iter()
         .map(|pkg| pkg.name())
         .collect::<Vec<_>>();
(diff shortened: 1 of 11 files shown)


Morganamilo:
I purposley left this out of paru because I think the way it was done in yay was pretty lacking. I also don't really see any need to tie it to an aur helper. I would like to see some one make a seperate package stats package that does this sort of thing in much more depth.

vojta7:
I would like to have this information available in paru. So I could try to dig into it, or are you planning to create this stats package yourself? Can you share with me a little bit more info about how do you imagine this crate, what kind of stats it should include, and how it would be used from paru?

vandah:
@Morganamilo I would also like to have this in paru. The absence of this is actually what's keeping me from switching from yay. Would it be possible at least as an opt-in feature?

Morganamilo:
I'll merge this if you also add the completion for it. I'd still prefer a more fleshed out tool rather than it being in an aur helper though.

src/stats.rs
```diff
@@ -0,0 +1,138 @@
...
+        if let Err(alpm::Error::PkgNotFound) = sync_db.pkg(pkg.name()) {
+            foreign_packages.push(pkg.name());
+        }
+        if pkg.reason() == PackageReason::Explicit {
+            explicit_packages += 1;
+        }
+        total_size += pkg.isize();
+    }
+
+    let aur_info = config.raur.info(&foreign_packages).await?;
```
Morganamilo, line 50: You can't just assume foreign package == aur package. They may be in a repo.

Morganamilo:
Do you plan on finishing this? All that needs to be done is fix this not accounting for local repos.

If not this implementation is almost done and if some one else wants to pick this up feel free.

src/stats.rs
vojta7: I am probably not understanding you correctly. I assume that aur packages are a subset of foreign packages, and this line (50), should get me a list of installed aur packages, is this the wrong assumption?

vojta7:
Hi, sorry for me taking so long to respond. I did rebase this branch onto master, but I am unsure how to resolve your feedback (Mentioned in the review comment).

src/stats.rs
Morganamilo: Yes packages may belong to a local repo. There's the two functions I mentioned in #357 but thinking about it neither of them fit perfectly. There's code in install.rs line 1684-1691 that does what you want. You should move that to a function in utils and use it. Following convention the name should be `repo_aur_pkgs` returning a list of repo and a list or aur packages.

vojta7: Thank you, it should now work as expected.

Morganamilo: I still see the code using a variable called foreign packages. I don't think the concept of foreign packages matters.

I think it should be:

```
Total packages:
Repo packages:
Aur packages:
```

(make sure to check the aur packages are actually in the aur)

Also with the odd and orphaned packages you should also add packages missing from the aur.

You can use `cache_info` to query all the packages into a hashmap then just iterate over the aur packages and see which ones are not in the hashmap.

There's also a Warnings struct in download.rs that you can use to do the warning printing for you.

Morganamilo: Also with `repo_aur_pkgs` my intention was for it to return repo and aur pkgs (hence the name). Seems as you wrote it to only return aur pkgs and the repo ones aint needed name it `aur_pkgs`.

vojta7: Yes, I misunderstood you, but It looks like I need both repo and aur packages, so I reworked the function as you intended it previously. To be completely honest, I am not really sure about this part of the code around repo packages, but I did my best to try to implement it correctly.

src/stats.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    print_line_separator(config);
+
+    let aur_packages = repo_aur_pkgs(config)
+        .1
+        .iter()
+        .map(|pkg| pkg.name())
+        .map(|s| s.to_owned())
+        .collect::<Vec<_>>();
```
vojta7, line 122: I don't really like querying information about aur packages in the middle of printing, but I didn't found any better way around borrow checker due to reference inside Warnings struct.

Morganamilo: You can create a new cache instead of using the one in config. The one in config should be refactored out really.

src/util.rs
```diff
@@ -327,3 +327,16 @@ pub fn split_repo_aur_pkgs<S: AsRef<str> + Clone>(config: &Config, pkgs: &[S]) -
...
+pub fn repo_aur_pkgs(config: &Config) -> (Vec<alpm::Package<'_>>, Vec<alpm::Package<'_>>) {
+    if config.repos != LocalRepos::None {
+        let (repo, aur) = repo::repo_aur_dbs(config);
+        let repo = repo.iter().flat_map(|db| db.pkgs()).collect::<Vec<_>>();
+        let aur = aur.iter().flat_map(|db| db.pkgs()).collect::<Vec<_>>();
+        (repo, aur)
+    } else {
+        let mut pkgs = config.alpm.localdb().pkgs().iter().collect::<Vec<_>>();
+        pkgs.retain(|pkg| config.alpm.syncdbs().pkg(pkg.name()).is_err());
+        (Vec::new(), pkgs)
```
Morganamilo, line 340: You need to fill in the repo packages here. In this case aur is the inverse of repo. So you can just do `config.alpm.localdb().pkgs().iter()` then filter with `config.alpm.syncdbs().pkg(pkg.name()).is_err()` for aur packages and `config.alpm.syncdbs().pkg(pkg.name()).is_ok()` for repo packages.

src/stats.rs
```diff
@@ -0,0 +1,135 @@
...
+    total_size: i64,
+    max_packages: Vec<(i64, &'a str)>,
+}
+
+async fn collect_info<'a>(config: &'a Config, max_n: usize) -> Result<Info<'a>> {
+    let db = config.alpm.localdb();
+    let sync_db = config.alpm.syncdbs();
+
+    let total_packages = db.pkgs().len();
+    let mut foreign_packages = Vec::new();
```
Morganamilo, line 28: Unused.

src/stats.rs
```diff
@@ -0,0 +1,135 @@
...
+        if let Err(alpm::Error::PkgNotFound) = sync_db.pkg(pkg.name()) {
+            foreign_packages.push(pkg.name());
+        }
+        if pkg.reason() == PackageReason::Explicit {
+            explicit_packages += 1;
+        }
+        total_size += pkg.isize();
+    }
+
+    let (repo, aur) = repo_aur_pkgs(config);
```
Morganamilo, line 48: aur is just the possible aur packages. When you do cache_info_with_warnings, warnings.pkgs.len() is the real amount of aur packages.

src/stats.rs
vojta7: Local cache helps me only partially, another pain point is a reference to pkgs inside Warnings struct. I left this part of the code where it was but replaced cache from config with a local variable, to help with possible future refactoring.

src/stats.rs
vojta7: Removed, I missed this.

Morganamilo:
Looks good to me now.

PR #259, opened 2021-02-22
repair broken build process

FoxieDev:
The build process is currently broken due to some sort of issue with the arch setup creating an entry for the keyserver. I've used the ubuntu keyserver as it seems the most performant from my testing, but feel free to modify this PR to change to a more preferred keyserver.

PR #262, opened 2021-02-23
Fix build pipeline

ejose19:


ejose19:
@Morganamilo can you trigger a pipeline to confirm if this fixes the issue? So #259 can cherry-pick the 2nd commit and merge.

Morganamilo:
You should be able to test this yourself by just pushing a tag starting with v.

Morganamilo:
I see a lot of extra stuff in this PR stuff that we were exploring in #259. Is that stuff actually needed for this to work or are you just  building on that?

ejose19:
Was waiting for pipeline (https://github.com/ejose19/paru/actions/runs/594038571) to finish to confirm you, but I managed to get it to work. Main issue is https://github.com/actions/virtual-environments/issues/2658, which first resolution was to use https://github.com/actions/virtual-environments/issues/2698#issuecomment-779262068, however that didn't work as seem gh is using an older host, so the other resolution was to use a patched version of glibc that works for linux4 (based on PR https://github.com/lxqt/lxqt-panel/pull/1562) which solves the issue for x86_64 (arm build doesn't seems to be affected).

Regarding the pacman.conf changes, yes they're needed. Seems github screwed another thing so these two must be kept until solved upstream.

Morganamilo:
I very much hate how awful the hacks are here but it works.

ejose19:
I agree with you, also I'm not very comfortable using an unsigned package in the base image of paru build. So either:
- Wait for an official package of glibc-linux4
- Wait for github to solve this (this started over 2 weeks, so who knows how long it will take)
- Migrate building to a more mature solution (travis, circleci)

I will close this PR for now.

ejose19:
@Morganamilo I've continue to investigate about the issue, seems that `buildx` is not fully compatible with the patch of `runc`, when I removed `buildx` the 1st workaround worked without doing anything extra, so I filed https://github.com/moby/buildkit/issues/1995 to monitor the progress of that particular issue.

Regarding the other issues (ro mounts), it's `buildx` again with https://github.com/moby/buildkit/issues/1267, and applying the workaround specified in comments allows to continue with the simplest change in `Dockerfile`.

So either we keep waiting for a proper solution, or we split the `build.yml` so `aarch64` is built with `buildx` (`aarch64` is not affected by glibc bug) and `x86_64` is built using `docker build` directly (on which the latest `runc` works and there won't be a need to edit `pacman.conf`)

Morganamilo:
The build worked for without the last commit, what does the last commit actually fix?

Morganamilo:
My hope is this will be fixed by the next release. At least some of the issuess. I used this PR to build 1.3.0, hoping the next release will be a while off.

ejose19:
This is still a work in progress, as I said I'm waiting for a reply on https://github.com/moby/buildkit/issues/1995 to finish this PR. The previous "commit" included the 2nd workaround but it was only a proof of concept to confirm the issue solved, I'm pretty sure it's not the solution we want and didn't thought you would publish that one.

So I'll mark this as a draft instead but wanted to share my advances on this with you. And I doubt this will be completely fixed by next release as buildx clearly stated they won't fix the readonly mounts, so at least that fix must be included in this PR.

ejose19:
The folks at `moby` already solved https://github.com/moby/buildkit/issues/1995, so this PR is complete and you can merge it (there's no need to use any patch now). Here's the [pipeline](https://github.com/ejose19/paru/actions/runs/598019044) to confirm it.

I would recommend releasing another version and deprecate/hide current latest one at that one was built with an unsigned package, and I'm sure paru users will appreciate using only trusted builds.

ejose19:
They already updated the default [moby/buildkit:buildx-stable-1](https://hub.docker.com/layers/moby/buildkit/buildx-stable-1/images/sha256-d8861bb3e7c0966156f728e68fcfd8a2ddb3fad30ffe4d39af35d409cbd46d67?context=explore) that comes with `
setup-buildx-action `, so I removed the reference to `latest`, leaving this PR with only the "essential" changes.

@Morganamilo Do you need anything else to be done here before merging?

Issue #254, opened 2021-02-19
Always review diffs


muhlinux:
### Option to always answer yes to `Proceed to review?`
If you always review the PKGBUILDs, it would be convenient to just print them instead of asking for confirmation.

Thanks for considering :space_invader:

muhlinux:
Somewhat related to #180

MTBorg:
I have a fix for this in #263.

Morganamilo:
Maybe if I changed the prompt to

### `Here are all the packages and deps that will be installed, as well as the conflicts if there are any. Would you like to continue on?

There would be less confusion.

PR #263, opened 2021-02-24
Allow user to always or never review

MTBorg:
This modifies the `SkipReview` option to allow the user to set it to `always`, `never` and `ask`.

If set to `ask` it will ask the user if it wants to skip the review process.
If set to `always` or `never` it will always or never show the review process, without ever prompting the user for input.

### DIFF
--- a/man/paru.conf.5
+++ b/man/paru.conf.5
@@ -204,8 +204,12 @@ Build packages in a chroot. This rquires the LocalRepo option to be enabled.
 Optionaly a directory may be passed to specify where the create the chroot.
 
 .TP
-.B SkipReview
+.B SkipReview [= always|never|ask]
 Skip the review process.
+If not set it defaults to ask.
+If empty or set to always it will always skip the review process without prompting the user for input.
+If set to never it will always perform the review process without prompting the user for input.
+If set to ask it will ask the user if it wants to skip the review process.
 
 .SH BIN
 Options belonging to the [bin] section.
--- a/src/config.rs
+++ b/src/config.rs
@@ -189,6 +189,8 @@ pub struct Config {
     #[default = "any"]
     pub mode: String,
     pub aur_filter: bool,
+    #[default = "ask"]
+    pub skip_review: String,
 
     #[default = 7]
     pub completion_interval: u64,
@@ -196,7 +198,6 @@ pub struct Config {
     pub help: bool,
     pub version: bool,
 
-    pub skip_review: bool,
     pub no_check: bool,
     pub no_confirm: bool,
     pub devel: bool,
@@ -636,12 +637,12 @@ impl Config {
             "makedepends",
             "optdepends",
         ];
+        let always_never_ask = &["always", "never", "ask"];
 
         let mut ok1 = true;
         let mut ok2 = true;
 
         match key {
-            "SkipReview" => self.skip_review = true,
             "BottomUp" => self.sort_mode = "bottomup".into(),
             "AurOnly" => self.mode = "aur".into(),
             "RepoOnly" => self.mode = "repo".into(),
@@ -686,6 +687,10 @@ impl Config {
                 let value = value.unwrap_or("yes").into();
                 self.remove_make = validate(value, yes_no_ask)?;
             }
+            "SkipReview" => {
+                let value = value.unwrap_or("always").into();
+                self.skip_review = validate(value, always_never_ask)?;
+            }
             "UpgradeMenu" => self.upgrade_menu = true,
             "LocalRepo" => self.repos = LocalRepos::new(value),
             "Chroot" => {
@@ -716,6 +721,7 @@ impl Config {
             "Redownload" => self.redownload = validate(value?, no_all)?,
             "Rebuild" => self.rebuild = validate(value?, no_all)?,
             "RemoveMake" => self.remove_make = validate(value?, yes_no_ask)?,
+            "SkipReview" => self.skip_review = validate(value?, always_never_ask)?,
             "SortBy" => self.sort_by = validate(value?, sort_by)?,
             "SearchBy" => self.search_by = validate(value?, search_by)?,
             "CompletionInterval" => self.completion_interval = value?.parse()?,
--- a/src/install.rs
+++ b/src/install.rs
@@ -10,7 +10,7 @@ use crate::keys::check_pgp_keys;
 use crate::print_error;
 use crate::repo;
 use crate::upgrade::get_upgrades;
-use crate::util::{ask, get_provider, split_repo_aur_targets, NumberMenu};
+use crate::util::{ask, ask_review, get_provider, split_repo_aur_targets, NumberMenu};
 use crate::{args, exec, news};
 
 use std::collections::hash_map::Entry;
@@ -448,20 +448,14 @@ async fn prepare_build<'a>(
         false
     };
 
-    if !config.skip_review {
-        if !ask(config, "Proceed to review?", true) {
-            return Ok(BuildInfo::stop());
-        }
-    } else {
-        if !ask(config, "Proceed with install?", true) {
-            return Ok(BuildInfo::stop());
-        }
+    if !ask(config, "Proceed with install?", true) {
+        return Ok(BuildInfo::stop());
     }
 
     let bases = Bases::from_iter(actions.iter_build_pkgs().map(|p| p.pkg.clone()));
     let srcinfos = download_pkgbuilds(config, &bases).await?;
 
-    if !config.skip_review {
+    if config.skip_review == "never" || (config.skip_review == "ask" && ask_review(config)) {
         let ret = review(config, &actions, &srcinfos, &bases)?;
         if ret != 0 {
             let mut bi = BuildInfo::stop();
--- a/man/paru.8
+++ b/man/paru.8
@@ -270,8 +270,11 @@ Sort AUR results by a specific field during search. Defaults to votes.
 Search for AUR packages by querying the specified field. Defaults to name-desc.
 
 .TP
-.B \-\-skipreview
+.B \-\-skipreview [= always|never|ask]
 Skip the review process.
+If empty or set to always it will always skip the review process without prompting the user for input.
+If set to never it will always perform the review process without prompting the user for input.
+If set to ask it will ask the user if it wants to skip the review process.
 
 .TP
 .B \-\-review
--- a/src/command_line.rs
+++ b/src/command_line.rs
@@ -141,6 +141,7 @@ impl Config {
             "makedepends",
             "optdepends",
         ];
+        let always_never_ask = &["always", "never", "ask"];
 
         match takes_value(arg) {
             TakesValue::Required if value.is_none() => bail!("option {} expects a value", arg),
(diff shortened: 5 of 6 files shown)


Morganamilo:
I think I need to decide on how exactly this should work. Adding the options to skipreview feels a bit awkward. Ideally the option would be `Review =` and by on in the default config. But it's a bit late for that as we can't alter the config for people already using paru. And I don't want to open up the possibility of accidental pkgbuild execution.

MTBorg:
Ok, i'll leave it open until you've decided. Can you elaborate on what you mean feels awkward?
I actually started implementing it as a boolean `Review` but chose to go with string `SkipReview` instead as i didn't feel it made sense to have two booleans (rather than a string/enum) controlling more or less the same behavior.

Also i realised that this won't show the diffs, even when set to `always`, when the package exists in the cache. I assume this is not the behavior one would want?

Morganamilo:
> Can you elaborate on what you mean feels awkward?

It's awkard because it's inverted.

> Also i realised that this won't show the diffs, even when set to always, when the package exists in the cache. I assume this is not the behavior one would want?

Well a diff is the differences between when you last saw it and now. So if there's nothing to show there's nothing to show. Even if set to always show diffs.

src/config.rs
```diff
@@ -686,6 +686,10 @@ impl Config {
                 let value = value.unwrap_or("yes").into();
                 self.remove_make = validate(value, yes_no_ask)?;
             }
+            "SkipReview" => {
+                let value = value.unwrap_or("ask").into();
```
Morganamilo, line 690: Also the default forr ask if very weird.

setting `SkipReview` without a value should be always.

Also ask is the default if you don't put anything so there's no sense in it being a valid option.

Just `SkipReview = Always|Never`

MTBorg: > It's awkard because it's inverted.

> setting SkipReview without a value should be always.

> Also ask is the default if you don't put anything so there's no sense in it being a valid option.
> Just SkipReview = Always|Never

I agree with these, should i go ahead and fix it?

> Well a diff is the differences between when you last saw it and now. So if there's nothing to show there's nothing to show. Even if set to always show diffs.

Yes, but then it feels somewhat misleading to call the option `always` as it's _not_ always. Maybe it should be called anything else? The documentation should atleast be updated to be clear about this.

MTBorg:
> It's awkard because it's inverted.

> setting SkipReview without a value should be always.

> Also ask is the default if you don't put anything so there's no sense in it being a valid option.
> Just SkipReview = Always|Never

I agree with these, should i go ahead and fix it?

> Well a diff is the differences between when you last saw it and now. So if there's nothing to show there's nothing to show. Even if set to always show diffs.

Yes, but then it feels somewhat misleading to call the option `always` as it's _not_ always. Maybe it should be called anything else? The documentation should atleast be updated to be clear about this.

Morganamilo:
> I agree with these, should i go ahead and fix it?

Yes, but not the but about it being inverted if you were including that too.


> Yes, but then it feels somewhat misleading to call the option always as it's not always. Maybe it should be called something else? The documentation should atleast be updated to be clear about this.

Makes sense to me. Especially since paru will print "nothing new to review" you you know it's not just broken.

MTBorg:
> Also ask is the default if you don't put anything so there's no sense in it being a valid option.
> Just SkipReview = Always|Never

Actually now that i come to think about it, why shouldn't it be a valid option? This is the same behavior as with other options, such as `SortBy` and `SearchBy`, in that you can set an option manually to its default value.

Morganamilo:
I do try to avoid the redundant options. There's no TopDown or NoDevel in the config because they are the defaults. Similar there's `Rebuild [= all]` without the option to do `Rebuild = No` or `Rebuild = Targets`

The reasons SortBy  has all the options is because there is no default. You can't just do `SortBy` in the config like how you can do with `NoReview`.

MTBorg:
That makes sense, i'll update the pull-request whenever i have time :)

MTBorg:
I did an update, to summarize:
- The option is called `SkipReview` 
- If not set it defaults to `ask`, to avoid accidental PKGBUILDs
- If empty or set to always it will always skip the review process (no user input needed).
- If set to never it will never skip the review process (no user input needed).
- If set to ask, it will always ask the user.

Is this what you had in mind?

src/install.rs
```diff
@@ -461,7 +461,7 @@ async fn prepare_build<'a>(
     let bases = Bases::from_iter(actions.iter_build_pkgs().map(|p| p.pkg.clone()));
     let srcinfos = download_pkgbuilds(config, &bases).await?;
 
-    if !config.skip_review {
+    if config.skip_review == "never" {
```
Morganamilo, line 464: Shouldn't this be `!= always`?

MTBorg: Yes, that it is true, good catch :) Fixed it.

man/paru.conf.5
```diff
@@ -204,8 +204,12 @@ Build packages in a chroot. This rquires the LocalRepo option to be enabled.
 Optionaly a directory may be passed to specify where the create the chroot.
 
 .TP
-.B SkipReview
+.B SkipReview [= ask] <always|never|ask>
```
Morganamilo, line 207: This should be:

.B SkipReview [= always|never|ask]

Morganamilo:
For consistency --skipreview should also take an opt arg. But that's it for my complaints.

man/paru.conf.5
MTBorg: Shouldn't `[= ask|always|never]`, i.e. what comes after `=` is the default when the option is not set?
This seems to be how the `RemoveMake` option is implemented/documented atleast.

```
RemoveMake [= ask]
```
```rust
"RemoveMake" => {
                let value = value.unwrap_or("yes").into();
                self.remove_make = validate(value, yes_no_ask)?;
}
```
```rust
#[default = "no"]
    pub remove_make: String,
```

Morganamilo: The first value is what the option is set to if set without a value.

MTBorg: Hmm, okay, but then either the documentation or implementation for `RemoveMake` incorrect, right?

MTBorg:
> For consistency --skipreview should also take an opt arg. But that's it for my complaints.

I assume you mean it `--skipreview` should be of the form 
```shell
--skipreview <yes|no>
```
and that i can remove the `--review` option and only use the `--skipreview`?

man/paru.conf.5
Morganamilo: No actually I just didn't think what I was saying. That's in the general case, but yes isn't an option because it's redundant.

Morganamilo:
the flag should mimic the config option. never always ask. I'm not sure about removing --review. May be good to keep it for compatibility.

MTBorg:
I've fixed the documentation and the cli option. Please have a look :)

src/install.rs
```diff
@@ -448,7 +448,7 @@ async fn prepare_build<'a>(
         false
     };
 
-    if !config.skip_review {
+    if config.skip_review == "ask" {
```
Morganamilo, line 451: This should be != never. But that's all I see wrong now.

MTBorg: If it's `!= never` then it will also be true for `always` but `always` shouldn't ask for user input.

Morganamilo: The double negative keeps getting me. != always in that cause. Basically the prompt should ask if you want to review you don't skip the review process.

MTBorg: I'm not quite sure i understand you.
That line should be `== "ask"` because only `ask` should prompt the user for input, both `always` and `never` should not prompt the user for input.

Morganamilo: Look at the else.

MTBorg: Can you elaborate? I see nothing wrong.

Morganamilo: All the if does is change the prompt message. If there's any chance to review (skip = ask or never) then the prompt should ask to continue to review.

MTBorg: Changing the prompt message is the same behavior as before. 

If the user sets `SkipReview` to `never` then they should not be prompted (otherwise `never` has the same the same effect as `ask`),  they should only be prompted for input when `SkipReview` is set to `ask`. That is the entire idea of implementing `always` and `never` for `SkipReview` in the first place, if the user always or never reviews diffs then they should not have to be prompted for input every time.

Morganamilo: The prompt is unrelated to review. It simply asks if you want to continue to the next step. It's an if-else that just changes the message. You're unconditionally prompted. The only issue with this PR the prompt message is wrong when skipreview = never.

Morganamilo: I thought there was an actual prompt to ask if you want to review but I guess I may have confused that with yay. In that case you'd be free to add one but given what you want is to always review I doubt you'd want that.

MTBorg: Ok, i think i see what you mean, please have a look at the new commit.

MTBorg: To summarize:
Regardless of `SkipReview`s value the same prompt after displaying conflicts, etc, asking the user to continue with the installation.
Then, if `SkipReview` is set to `ask`, the user is prompted for input whether or not the review the diffs.
If the user answered yes or if `SkipReview` is set to `never` then the review is shown.
If `SkipReview` is set to `always` the user is not prompted and the diffs are not shown.

MTBorg: I added the `ask_review` function as i think if `SkipReview` is set to ask then `--no-confirm` should not have any effect on the prompt.

Morganamilo: So what you really want is just the ability to use --noconfirm and also review?

MTBorg: Essentialy yes, that is the behavior i want. See #254.

Morganamilo: #254 doesn't ask for the behaviour you're describing. If you just want --noconfirm with pause to confirm the review I think this isn't really the write aproch. --noconfirm needs to not prompt for the sake of scripting. I think a dedicated flag would be better than chucking it in with skip review.

MTBorg: Hmm, maybe i wasn't clear. Like i said, that's the behavior i want but it's not the use case i want. I just meant that using `--noconfirm` and getting to review, would produce the same result as what i wanted for the review process. But using `--noconfirm` would also affect the rest of the entire installation process.
This is unrelated to `--noconfirm`, that was just something i had to workaround because how `ask` (the function) was implemented. 
Personally i don't use `--no-confirm`, i just don't want to be prompted whether or not review everytime, because i always do review.

Morganamilo: Okay, but as is, Paru does not ask if you want to review it does automatically.

MTBorg: Are you saying that paru does not ask the user if they want to review? Because that is not the behavior that i'm seeing when installing packages
```
# paru tealdeer
1 aur/tealdeer 1.4.1-5 [+42 ~3.11] [Installed]
    A fast tldr client in Rust.
2 aur/tealdeer-git r255.30b7c5f-1 [+10 ~0.00]
    A fast tldr client in Rust.
3 aur/tealdeer-bin 1.4.1-1 [+6 ~0.61]
    A fast tldr client in Rust. Static binaries from GitHub.
:: Packages to install (eg: 1 2 3, 1-3):
:: 2
:: Resolving dependencies...
:: Calculating conflicts...
:: Calculating inner conflicts...

:: Conflicts found:
    tealdeer-git: tealdeer (tldr)  tealdeer (tldr)

:: Conflicting packages will have to be confirmed manually

Aur (1) tealdeer-git-r255.30b7c5f-1

:: Proceed to review? [Y/n]:
```

Morganamilo: Yes, paru tell you what it's going to install then asks if you're okay with that and want to continue to the next step. It is nothing to do with review.

Morganamilo: Not to mention after this prompt pkgbuilds are downloaded. So the prompt is also there to stop you from just downloading pkg builds accidentally.

MTBorg: Ah, yes i see what you mean now. I guess this pr is redudant then. Should i close it?

jpegxguy:
This was close but I liked the idea. I essentially want to always review after having chosen which packages I want to install.
I'd like the prompt for review to not exist, and just act like i pressed Yes

PR #266, opened 2021-02-27
Add an option to show upstream URLs

PillTime:
Added an option so that, when looking for packages to install, they show their upstream URL.

Whenever I'm looking for a package, I like to see if they're updated regularly. So I go to the source to see when the last update was.

This option takes way the burden of doing `-Si` for every package. And since it's optional, if people don't want this, they can keep it turned off.

### DIFF
--- a/man/paru.8
+++ b/man/paru.8
@@ -397,6 +397,10 @@ gpg config\%.
 Print new news during sysupgrade.
 
 .TP
+.B \-\-showupstream
+Print the upstream url of packages.
+
+.TP
 .B \-\-useask
 Use pacman's --ask flag to automatically confirm package conflicts. Paru lists
 conflicts ahead of time. It is possible that paru does not detect a conflict, 
--- a/man/paru.8
+++ b/man/paru.8
@@ -72,8 +72,8 @@ built packages but will keep already downloaded VCS sources.
 
 .TP
 .B \-Sss
-Paru will output a verbose search result when passing two '--search' or '-s'
-flags, ultimately showing the upstream URL of each package.
+Paru will also output a verbose search result when passing two
+'--search' or '-s' flags.
 
 .TP
 .B \-S, \-Si, \-Sl, \-Ss, \-Su, \-Qu
--- a/man/paru.8
+++ b/man/paru.8
@@ -71,6 +71,11 @@ cache. Cleaning untracked files will wipe any downloaded sources or
 built packages but will keep already downloaded VCS sources.
 
 .TP
+.B \-Sss
+Paru will output a verbose search result when passing two '--search' or '-s'
+flags, ultimately showing the upstream URL of each package.
+
+.TP
 .B \-S, \-Si, \-Sl, \-Ss, \-Su, \-Qu
 Paru will also handle AUR packages for these operations.
 
@@ -395,10 +400,6 @@ gpg config\%.
 .TP
 .B \-\-newsonupgrade
 Print new news during sysupgrade.
-
-.TP
-.B \-\-showupstream
-Print the upstream url of packages.
 
 .TP
 .B \-\-useask
--- a/src/search.rs
+++ b/src/search.rs
@@ -170,14 +170,14 @@ fn print_pkg(config: &Config, pkg: &raur::Package, quiet: bool) {
         .split_whitespace();
     print_indent(Style::new(), 4, 4, config.cols, " ", desc);
 
-    if config.show_upstream && pkg.url.is_some() {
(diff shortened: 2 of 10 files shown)


Morganamilo:
How about -Sss as a verbose search? Kind of to match -Sii. I'm also not a fan of the formatting I won't lie.

PillTime:
I like the idea of a verbose search.
As for the formatting, I didn't really know where to put the URL. I thought of putting it at the end of the line with the package name, but I feel like that would look worse.

Being a verbose search, I guess putting some more info like the license would make sense?

Morganamilo:
There doesn't really have to be more info. -Sii only adds a few extra fields.

Formatting I don't know, but at the very least `Upstream:` should be bold. Also it should probably be `Url:` to match the field name.

PillTime:
A couple of questions.

First, URL before or after the decription?


Second, if we were to put this in a verbose search command like `-Sss`, would it still be possible to have some way of showing the URL in the interactive menu (when doing `paru <pkg>`)?

Morganamilo:
> First, URL before or after the decription?

Not sure on the formatting.

> Second, if we were to put this in a verbose search command like -Sss, would it still be possible to have some way of showing the URL in the interactive menu (when doing paru <pkg>)?

`paru -ss foo` would work.

PillTime:
Would it be worth it to put some information on the man page about the double search flag, since it's something `pacman` doesn't have?

PillTime:
Alright, using `-Sss` now, and I added an entry in the man page since it's something `pacman` doesn't have.

src/search.rs
```diff
@@ -169,6 +169,18 @@ fn print_pkg(config: &Config, pkg: &raur::Package, quiet: bool) {
...
     print_indent(Style::new(), 4, 4, config.cols, " ", desc);
+
+    if config.pacman_args().count("s", "search") > 1 && pkg.url.is_some() {
+        print!("    ");
+        let upstream_url = pkg
+            .url
+            .as_deref()
+            .unwrap() // is Some(), can't fail
+            .to_owned();
+        let upstream = format!("{} {}", Style::new().bold().paint("URL:"), upstream_url);
```
Morganamilo, line 180: use c.bold

src/search.rs
```diff
@@ -169,6 +169,18 @@ fn print_pkg(config: &Config, pkg: &raur::Package, quiet: bool) {
         .unwrap_or("None")
         .split_whitespace();
     print_indent(Style::new(), 4, 4, config.cols, " ", desc);
+
+    if config.pacman_args().count("s", "search") > 1 && pkg.url.is_some() {
```
Morganamilo, line 173: I'd prefer `if let Some(ref url) = pkg.url { if config.pacman_args()...``

Morganamilo, line 173: use `config.args` over `config.pacman_args()`

man/paru.8
```diff
@@ -70,6 +70,11 @@ Paru will also clean cached AUR packages and any untracked files in the
 cache. Cleaning untracked files will wipe any downloaded sources or
 built packages but will keep already downloaded VCS sources.
 
+.TP
+.B \-Sss
```
Morganamilo, line 74: I'd rather just say it gives some extra verbose output. I feel what -Sss does may get added to over time so may as well keep it generic.

Morganamilo (commented):
I'm on board with -Sss, though looking at the output I'd go for description last.

Also would be nice if you can also add AUR URL (see paru -Si) as I tend to use that link a lot.

PillTime:
Would it make sense for the description to also get a tag like **`Description:`** at its start, since the other fields all have one? To me it looks like we're giving more importance to the URLs (with them having a bold tag at the beginning) when they should all be equal.

Morganamilo:
I'm not sure. I think as description is the thing in the bottom it should be easy to snap your focus too. Plus description is the most likely to be long so maybe prefixing it with Description: would take too much space away from it.

Also thinking about it, you can replace your print logic with `info::print()`. With that it should be good.

PillTime:
> I think as description is the thing in the bottom it should be easy to snap your focus too.

Fair enough.

> you can replace your print logic with `info::print()`.

Oh wow, yeah. Looks much better.

Morganamilo:
I did some playing around with the formatting. Made some tweaks and merged. Ended up settling with the desc on top. Tell me if you have anything against the formatting.

PillTime:
Ahh, didn't think of that.
I like it. The description doesn't go so unnoticed thanks to the "empty" space below it.
I'm alright with this.

Morganamilo:
Alright then. Thanks for the PR.

Issue #296, opened 2021-03-29
Paru failing to run: pacman: No such file or directory (os error 2) 


famfo:
### Affected Version
`paru -V`: paru v1.4.0 +backtrace - libalpm v12.0.2

### Description
Paru failes to run pacman

### Output
`paru -Suy`
`error: failed to run: pacman --sync -u -y --: No such file or directory (os error 2)`

trevorgunn:
Do you have sudo installed?

trevorgunn:
Aliases only work in your shell. Paru needs an actual program.

There is a config option in paru.conf for this. Make sure to uncomment both `[bin]` and `Sudo = doas`.

famfo:
This fixed it, than you for your help.

Issue #206, opened 2021-02-03
Paru error


PogMaster69:
Sorry for the unhelpful title I could not think of what to title it. `error: unknown option 'Sudo' in section [options]
error: failed to run: pacman --sync -y -u --: No such file or directory (os error 2)` is the error I get when just typing `paru` I believe its
a doas problem because when I type `doas pacman --sync -y -u` it works just fine. In my paru config I have uncommented `Sudo = doas`

edit: I also have sudo and doas alias in my zsh.rc folder. Also I'm new to linux and github so please don't roast me too hard

PogMaster69:
I had to include path to doas instead of just writing doas ;-;

PogMaster69:
never mind I'm getting a new error yay. `error: Password: doas: pam_authenticate: Permission denied`

Morganamilo:
> error: unknown option 'Sudo' in section [options] error:

You need to also uncomment `[bin]`


> never mind I'm getting a new error yay. error: Password: doas: pam_authenticate: Permission denied

I've never used doas so you're on your own with that one.

PogMaster69:
yeah I uncommented bin that resolves the first error but I'm still getting the second error on *some* downloads

Morganamilo:
The second does not seem paru related

triis:
I get a similar error where, "doas: invalid option -- 'v'
usage: doas [-Lns] [-C config] [-u user] command [args]
doas: invalid option -- 'v'
usage: doas [-Lns] [-C config] [-u user] command [args]" is printed at least twice before being prompted for my password. Could this be to do with the SudoLoop option?

Morganamilo:
Yes, sudo loop only works with sudo.

PR #297, opened 2021-03-29
Mention sudo_bin if an error occurs calling pacman

trevorgunn:
Hopefully this will help some users diagnose their own issues better such as in #206 and #296.

### DIFF
--- a/src/exec.rs
+++ b/src/exec.rs
@@ -94,23 +94,31 @@ pub fn pacman<S: AsRef<str> + Display + std::fmt::Debug>(
     config: &Config,
     args: &Args<S>,
 ) -> Result<Status> {
-    let mut command = if config.need_root {
-        let mut command = Command::new(&config.sudo_bin);
-        command.args(&config.sudo_flags);
-        command.arg(args.bin.as_ref());
-        command
-    } else {
-        Command::new(args.bin.as_ref())
-    };
+    let mut command: std::process::Command;
+    let error_msg: String;
 
     if config.need_root {
         wait_for_lock(config);
+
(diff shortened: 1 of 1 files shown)


src/exec.rs
```diff
@@ -94,23 +94,30 @@ pub fn pacman<S: AsRef<str> + Display + std::fmt::Debug>(
...
+    let mut command: std::process::Command;
+    let error_msg: String;
 
     if config.need_root {
         wait_for_lock(config);
+
+        command = Command::new(&config.sudo_bin);
+        command.args(&config.sudo_flags);
+        command.arg(args.bin.as_ref());
+        error_msg = format!("failed to run `{}` with arguments: {} {} {}",
```
Morganamilo, line 106: I don't see any need for the `with arguments` bit. Also this causes excess spaces when sudoflags is empty (which it usually will be).

trevorgunn: Right, I had thought about adding dealing with the extra space but I didn't feel like adding another 4 lines of code and an extra if/else statement just to deal with one space.

But you're right, just forget the arguments since this error will (I assume) only appear if the first program can't be spawned. Otherwise it will be up to that program to provide a suitable error message.

I will make the changes.

Morganamilo:
No i'd prefer the flags to be in there just in case. I don't really care about the spaces if it's that much effort. The error shouldn't come up very often anyway.

trevorgunn:
Do you want the pacman command in there as well? Or just the flags for sudo?

Morganamilo:
All of the flags to whatever was ran.

PR #303, opened 2021-04-01
Add Makefile

exorcist365:
Added a Makefile, for easier installation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,29 @@
+DESTDIR = /usr/local
+BINDIR = $(DESTDIR)/bin
+MANDIR = $(DESTDIR)/share/man
+
+paru:
+	@set -o errexit -o pipefail
+	@cargo build --release --locked --target-dir target
+	@strip target/release/paru
+	@tar --zstd -cfparu.tar.zst  man completions paru.conf -C target/release paru
+
+clean:
+	@cargo clean
+
+install: paru
+	@mkdir -pv $(BINDIR)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -0,0 +1,29 @@
+DESTDIR = /usr/local
```
E5ten, line 1: This should be PREFIX, not DESTDIR, BINDIR and MANDIR should use PREFIX instead of DESTDIR, and then $(DESTDIR) should be prepended to $(BINDIR) and $(MANDIR) when they're used in targets, so DESTDIR is still used even if the user manually sets BINDIR or MANDIR to something else

Makefile
```diff
@@ -0,0 +1,29 @@
...
+BINDIR = $(DESTDIR)/bin
+MANDIR = $(DESTDIR)/share/man
+
+paru:
+	@set -o errexit -o pipefail
```
E5ten, line 6: This is not needed, as make will exit if any of the commands fail anyway, and there are no pipes for pipefail to affect. Additionally, pipefail is bash-specific and may not work if the shell being used by make is not bash. Finally, depending on implementation, each command in the target might run in its own shell, so the set call might not affect the following calls at all.

Morganamilo:
I don't see any point in a makefile, It just throws another build system in to the mix and complicates things. Not to mention make isn't really suited to building the project as it can't configure the build process.

exorcist365:
Alright.

PR #307, opened 2021-04-03
Use makepkg to generate srcinfo with paru -U

### LINKED ISSUE #305: Automatically generate .SRCINFO when installing local PKGBUILD
When installing local PKGBUILD with `paru -U` in a directory without `.SRCINFO` the installation fails

I think something like `makepkg --printsrcinfo > .SRCINFO` should be done automatically by paru

trevorgunn:
Resolve #305


Morganamilo:
I don't see a need to prompt or generate the file. Just parse the output of --printsrcinfo directly without saving it.

src/install.rs
```diff
@@ -96,7 +96,15 @@ pub async fn build_pkgbuild(config: &mut Config) -> Result<i32> {
     let arch = config.alpm.arch();
     let dir = std::env::current_dir()?;
     let srcinfo_dir = dir.join(".SRCINFO");
-    let srcinfo = Srcinfo::parse_file(&srcinfo_dir).context("failed to parse .SRCINFO")?;
+    let srcinfo_path = Path::new(&srcinfo_dir);
+
+    // Parse srcinfo from file, else from `makepkg --printsrcinfo`
+    let srcinfo = if srcinfo_path.exists() {
```
Morganamilo, line 102: I think we should go a step further an always just call --printsrcinfo. Otherwise, if there is for some reason an existing .srcinfo file and the user edits the pkgbuild these changes won't be reflected.

trevorgunn: So `makepkg` sources the PKGBUILD in order to generate the srcinfo. Usually this is fine because usually it is the PKGBUILD author who is generating the .SRCINFO file. And, if the user is running `paru -U` there is an expectation that they've checked the PKGBUILD so this is OK in this specific context especially given that makepkg will source the PKGBUILD a few lines later.

**HOWEVER** paru doesn't check if the UID is 0 until _after_ this step, so if the user has disabled that safeguard in `makepkg` then things could go very very bad.

So I'll make this change because I agree about the user possibly editing the PKGBUILD but I will throw in an extra check of the UID just to be on the safe side.

Morganamilo: Yeah the lack of root check is not something intentional.

Morganamilo:
Looks good now.

PR #315, opened 2021-04-08
Git is not in the base installation

dylanmtaylor:
Make sure that git is installed before attempting to run Git.


dylanmtaylor:
I know that this is a bit nitpicky, but being able to just copy/paste this block in on a fresh Arch install is really nice, and it's annoying to get nagged to install git.

trevorgunn:
#114

PR #324, opened 2021-04-14
Need to make $pkgdir readable, not writeable

### LINKED ISSUE #318: Paru -Scc permission denied
### Affected Version
**paru -V**
paru v1.5.0 - libalpm v12.0.2
### Description
When running paru -Scc it says Permission denied, even though I have the permissions. The only thing in the cache is darling which didn't build so I wanted to delete it from the cache.
### Output
**Include the FULL output of any relevant commands/configs**

```sh
 ~  paru -Scc

Cache directory: /var/cache/pacman/pkg/
:: Do you want to remove ALL files from cache? [y/N] y
removing all files from cache...

Database directory: /var/lib/pacman/
:: Do you want to remove unused repositories? [Y/n] y
removing unused sync repositories...

Clone Directory: /home/paulschadauer/.cache/paru/clone
:: Do you want to remove ALL AUR packages from cache? [y/N]: y
error: could not remove '/home/paulschadauer/.cache/paru/clone/darling': Permission denied (os error 13)

Diff Directory: /home/paulschadauer/.cache/paru/diff
:: Do you want to remove all saved diffs? [Y/n]: y
```

trevorgunn:
Fixes #318

### DIFF
--- a/src/clean.rs
+++ b/src/clean.rs
@@ -6,6 +6,7 @@ use crate::util::ask;
 
 use std::fs::{read_dir, remove_dir_all, remove_file, set_permissions, DirEntry};
 
+use std::os::unix::fs::PermissionsExt;
 use std::path::Path;
 use std::process::Command;
 
@@ -96,10 +97,10 @@ fn clean_aur(
     Ok(())
 }
 
-fn make_writeable(file: &Path) -> Result<()> {
+fn fix_perms(file: &Path) -> Result<()> {
     let pkg = file.join("pkg");
     let mut perms = pkg.metadata()?.permissions();
-    perms.set_readonly(false);
+    perms.set_mode(0o755);
     set_permissions(pkg, perms)?;
     Ok(())
 }
@@ -120,7 +121,7 @@ fn clean_aur_pkg(
         return Ok(());
     }
 
-    let _ = make_writeable(&file.path());
+    let _ = fix_perms(&file.path());
 
     if remove_all {
         remove_dir_all(file.path())


Morganamilo:
I don't really like the look of this library. It seems like it will fail on non Unicode paths and it also grows the stack. I don't think it would be very hard to do it yourself. Especially if using walkdir. Or alternatively just shelling out to rm.

trevorgunn:
Ok, I will try to work on a better solution later today.

Morganamilo:
The way I see it. Simply iterating with walkdir and setting readonly to false for each file should work. Then just remove_dir_all like usual.

trevorgunn:
So using Walkdir does not work because Walkdir does not build its iterator lazily. Meaning if the head of the iterator is currently pointing at directory D and you change the permissions of D to allow read-access you will still get an error calling `iter.next()` because Walkdir has already computed that there is an `io::Error` trying to read that directory.

This is why the rm_rf crate uses a recursive call. And, without lazy evaluation, I think a recursive call is really necessary here.  Unfortunately recursion means that the stack might blow up if directories are nested too deep.

I could maybe copy the rm_rf crate code and remove the problematic `.as_os_str().to_str()` and `stacker::maybe_grow` (I can't imagine any AUR package needing directories nested 1000 levels deep).

I still prefer to do it in Rust rather than calling `rm -rf` for better error handling, but certainly `std::process::Command::new("/usr/bin/rm")` is the easier solution.

trevorgunn:
This should resolve everything now.

1. All strings are kept as `OsStr`s
2. There's no recursive call so we don't need to worry about growing the memory stack. Instead we allocate in the heap.
3. There is room to add more detailed error messages (which would be impossible if we passed to `/usr/bin/rm -rf`.

Morganamilo:
While this may work, I think the real issue is the current code sets readonly to false but does not actually enable read perms. And given the error in the issue just seems to point to /pkg and makepkg removes read and write perms to /pkg during build I think there's the issue.

trevorgunn:
I follow you. So all we need to do is change the `make_writable` function to a `make_readable` function using `std::os::unix::fs::PermissionsExt` and then `remove_dir_all` should work as intended.

Maybe I'll just keep this approach in mind in case there is an error in the future with some package creating a non-readable directory.

src/clean.rs
```diff
@@ -96,10 +97,10 @@ fn clean_aur(
     Ok(())
 }
 
-fn make_writeable(file: &Path) -> Result<()> {
+fn make_readable(file: &Path) -> Result<()> {
     let pkg = file.join("pkg");
     let mut perms = pkg.metadata()?.permissions();
-    perms.set_readonly(false);
+    perms.set_mode(0o755);
```
Morganamilo, line 103: Directories should be 644

src/clean.rs
```diff
@@ -96,10 +97,10 @@ fn clean_aur(
     Ok(())
 }
 
-fn make_writeable(file: &Path) -> Result<()> {
+fn make_readable(file: &Path) -> Result<()> {
```
Morganamilo, line 100: Fix_perms probably makes more sense now.

src/clean.rs
trevorgunn: Files are 644, directories are 755 by default. The execute permission enables one to `cd` into that directory.

Although, it shouldn't make a difference either way since it will be removed right after.

Morganamilo: Yeah that's just me being dumb.

PR #328, opened 2021-04-15
Use enums instead of strings for most values in the Config struct.

### LINKED ISSUE #326: feat: paru could use enums instead of strings in its Config struct
The current approach is error-prone and involves a fair amount of code duplication.

KerfuffleV2:
This pull converts most of the string values in the `Config` struct to use enums instead and introduces a trait (`ConfigEnum`) for unified parsing/validating. 

I tried hard to change as little as possible and leave all semantics intact. There's one intended change in behavior as a result of this pull: Previously the config parsing code would accept only `all` or `no` values for some options from the config file while commandline processing allowed `all`, `yes` and `no`. This change will allow `yes` in the config also. If there's a reason to forbid setting `yes`  then I can make an adjustment, but I wasn't even sure the restriction was intentional.

So why is this approach preferable to the current one?

1. The typechecker will protect you against many issues when dealing with those values while its assistance is limited when manipulating string-based option values.
2. It unifies the processing of options and arguments making them easier to maintain and reason about.
3. Less code duplication.
4. Not really significant, but comparing enum values should be faster than strings.

**Note**: Only lightly tested at this point, although as far as I can see I didn't break anything. Most of the changes were straightforward.

Closes #326

### DIFF
--- a/src/clean.rs
+++ b/src/clean.rs
@@ -1,4 +1,4 @@
-use crate::config::Config;
+use crate::config::{Config, CfgMode};
 
 use crate::exec;
 use crate::print_error;
@@ -15,11 +15,11 @@ use anyhow::{bail, Context, Result};
 use srcinfo::Srcinfo;
 
 pub fn clean(config: &Config) -> Result<()> {
-    if config.mode != "aur" {
+    if config.mode != CfgMode::Aur {
         exec::pacman(config, &config.args)?;
     }
 
-    if config.mode != "repo" {
+    if config.mode != CfgMode::Repo {
         let remove_all = config.delete;
         let clean_method = &config.pacman.clean_method;
         let keep_installed = clean_method.iter().any(|a| a == "KeepInstalled");
@@ -31,7 +31,7 @@ pub fn clean(config: &Config) -> Result<()> {
             "Do you want to remove all other AUR packages from cache?"
         };
 
-        if config.mode == "any" {
+        if config.mode == CfgMode::Any {
             println!();
         }
 
--- a/src/clean.rs
+++ b/src/clean.rs
@@ -1,4 +1,4 @@
-use crate::config::{Config, CfgMode};
+use crate::config::{Config, Mode};
 
 use crate::exec;
 use crate::print_error;
@@ -15,11 +15,11 @@ use anyhow::{bail, Context, Result};
 use srcinfo::Srcinfo;
 
 pub fn clean(config: &Config) -> Result<()> {
-    if config.mode != CfgMode::Aur {
+    if config.mode != Mode::Aur {
         exec::pacman(config, &config.args)?;
     }
 
-    if config.mode != CfgMode::Repo {
+    if config.mode != Mode::Repo {
         let remove_all = config.delete;
         let clean_method = &config.pacman.clean_method;
         let keep_installed = clean_method.iter().any(|a| a == "KeepInstalled");
@@ -31,7 +31,7 @@ pub fn clean(config: &Config) -> Result<()> {
             "Do you want to remove all other AUR packages from cache?"
         };
 
-        if config.mode == CfgMode::Any {
+        if config.mode == Mode::Any {
             println!();
         }
 
--- a/src/command_line.rs
+++ b/src/command_line.rs
@@ -1,9 +1,11 @@
 use crate::args::{PACMAN_FLAGS, PACMAN_GLOBALS};
-use crate::config::{Colors, Config, LocalRepos};
+use crate::config::{
+    CfgMode, CfgOp, CfgSortMode, CfgValue, CfgYesNoAll, CfgYesNoAsk, Colors, Config, LocalRepos,
+};
 
 use std::fmt;
 
-use anyhow::{anyhow, bail, ensure, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use url::Url;
 
 #[derive(Debug, Copy, Clone)]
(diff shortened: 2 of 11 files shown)


Morganamilo:
Code doesn't compile:

```
error: lifetime may not live long enough
   --> src/devel.rs:345:22
    |
341 |   pub async fn pkg_has_update<'a>(
    |                               -- lifetime `'a` defined here
...
344 |       info: &HashSet<RepoInfo>,
    |             - let's call the lifetime of this reference `'1`
345 |   ) -> Option<&'a str> {
    |  ______________________^
346 | |     if info.is_empty() {
347 | |         return None;
348 | |     }
...   |
360 | |     }
361 | | }
    | |_^ function was supposed to return data with lifetime `'a` but it is returning data with lifetime `'1`

error: lifetime may not live long enough
   --> src/devel.rs:357:9
    |
341 | pub async fn pkg_has_update<'a>(
    |                             -- lifetime `'a` defined here
...
344 |     info: &HashSet<RepoInfo>,
    |           - let's call the lifetime of this reference `'1`
...
357 |         Some(pkg)
    |         ^^^^^^^^^ function was supposed to return data with lifetime `'1` but it is returning data with lifetime `'a`

error: aborting due to 2 previous errors

error: could not compile `paru`
```

KerfuffleV2:
Try `cargo clean` before trying to compile. That's a weird issue - I saw it too but cleaning the build directory always fixed it which seems to indicate the actual source isn't at fault.

Perhaps caused by some files that are dynamically generated in the paru build process? I'm not familiar enough with teh project to know if that could actually be the cause.

KerfuffleV2:
@Morganamilo I think this is interacting somehow with type/lifetime inference you're relying on but I really don't understand why it only happens intermittently. I've been trying to reproduce the issue without any success. Clippy doesn't like the current signature of that function though:

```
warning: explicit lifetimes given in parameter types where they could be elided (or replaced with `'_` if needed by type declaration)
   --> src/install.rs:428:1
    |
428 | / async fn prepare_build<'a>(
429 | |     config: &Config,
430 | |     aur_repos: HashMap<String, String>,
431 | |     cache: &mut Cache,
432 | |     mut actions: Actions<'a>,
433 | |     srcinfo: Option<&Srcinfo>,
434 | | ) -> Result<BuildInfo> {
    | |______________________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_lifetimes
```

However, they can't actually be removed or substituted with `'_`. Giving everything explicit lifetimes compiles and also shuts Clippy up:

```rust
pub async fn pkg_has_update<'pkg, 'conf, 'inf>(
    config: &'conf Config,
    pkg: &'pkg str,
    info: &'inf HashSet<RepoInfo>,
) -> Option<&'pkg str> {
```

I don't know if this actually helps with the other problem though, since I'm unable to reliably trigger it. Are you?

Morganamilo:
I made some changes here https://github.com/Morganamilo/paru/commit/6d34618c5e426dbd5766fe50a70251cf73f64b00.  Thought it would be faster to do it than make 100 comments. Tell me what you think.

KerfuffleV2:
> I made some changes here [6d34618](https://github.com/Morganamilo/paru/commit/6d34618c5e426dbd5766fe50a70251cf73f64b00). Thought it would be faster to do it than make 100 comments. Tell me what you think.

Naming convention is definitely personal taste, so I'll defer to your preference there. I'd say your function name changes are an objective improvement - I'm not that great at naming things in the "Rusty" way.

I would change [this](https://github.com/Morganamilo/paru/commit/6d34618c5e426dbd5766fe50a70251cf73f64b00#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecdR153) to at least use `.expect()` with a helpful error instead of unwrap. It's not quite an impossible error - it can occur if the enum name/value mapping tables are set up incorrectly. 

Is [this](https://github.com/Morganamilo/paru/commit/6d34618c5e426dbd5766fe50a70251cf73f64b00#diff-db0b940ca2fee0cb263f4ef9a75c23e4f77b8edb65e0b7ecec6d3f04ea4eae88R75) actually doing anything? If the types are just the same now, maybe the whole match can go away.

Other than those two things, no complaints about your modifications.

Morganamilo:
> I would change this to at least use .expect() with a helpful error instead of unwrap. It's not quite an impossible error - it can occur if the enum name/value mapping tables are set up incorrectly.

I prefer to use unwrap. If for some reason this ever did trigger I'd debug by the stacktrace so I don't see the need for a panic message. And if you think the error message is important for the user then go all the way and return a result with a proper error message.

 > Is this actually doing anything? If the types are just the same now, maybe the whole match can go away. edit: It does seem that you can just do let by = config.search_by; as all the cases appear to be covered, unless I'm blind.

Yeah didn't notice as the editor renamed it all. If you merge in my commit and make that change I'll then merge this.

KerfuffleV2:
> I prefer to use unwrap. 

I don't feel strongly enough to make the whole function chain deal with `Result`. `.expect()` still panics with a stack trace (as far as I know) just the output can be slightly more helpful. It's up to you though.

> If you merge in my commit and make that change I'll then merge this.

Done - I hope. I've never had to merge in upstream changes into my forks before so please let me know if I did something wrong.

Offtopic - there are a lot of Clippy warnings for the project. Would you accept a pull that applies the Clippy suggestions?

Morganamilo:
Sure, there is a pr on it but it's dead. However there are some clippy lints that I believe makes the code less readable in certain cases.

KerfuffleV2:
> However there are some clippy lints that I believe makes the code less readable in certain cases.

Well, the alternative is just adding a pragma to shut Clippy up. I'll open an issue to continue discussion on that topic.

PR #331, opened 2021-04-15
Give Clippy the world.

### LINKED ISSUE #330: Apply Clippy suggestions.
This is what it looks like giving Clippy what it wants: https://github.com/KerfuffleV2/paru/commit/fd61d005cf5b447a0aa27a3f2ccc97302a49434a

There are two cases which I didn't fix - `install.rs:build_install_pkgbuild()` and `upgrade.rs:print_upgrade()`. Both are functions that take many arguments but there didn't seem like a clear way to improve them. For those, I added a pragma to to make Clippy ignore them.

I think in general the changes were a modest improvement. If you don't hate it, I can submit that as a pull.

KerfuffleV2:
Closes #330

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -442,7 +442,10 @@ impl Ini for Config {
 
     fn callback(&mut self, cb: Callback) -> Result<(), Self::Err> {
         let err = match cb.kind {
-            CallbackKind::Section(section) => self.parse_section(section),
+            CallbackKind::Section(section) => {
+                self.section = Some(section.to_string());
+                Ok(())
+            }
             CallbackKind::Directive(_, key, value) => self.parse_directive(key, value),
         };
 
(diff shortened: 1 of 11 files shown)


Morganamilo:
The overly verbose lifetimes are annoying, but I also couldn't find a way to make clippy happy with the way it was.

Code needs to be ran through rustfmt then it's good.

KerfuffleV2:
Do you mean the longer names like `'conf`? I prefer that to arbitrary 1 letter ones when there are 3 or more lifetimes in play but I can change it if you'd like. 

I thought I already had run `rustfmt` on everything but I apparently missed a couple files. Changes were very minor: 2 sorted import blocks and one extraneous space removed.

Morganamilo:
I mean how clippy complains with how it was with one lifetime elided. Nothing actionable just annoying.

KerfuffleV2:
Yeah, I think it was a bit confused by the macros there. I'm not sure there is a real solution other than giving in or disabling the warning with a pragma and since I'm the pushover type...

Morganamilo:
I think it may actually be the async that confuses it. Given references can't be held between await points the lifetimes have to be there.

PR #351, opened 2021-05-01
Update deps

jimman2003:
Just run cargo update


Morganamilo:
Can you also upgrade the deps?

jimman2003:
Sure thing..

jimman2003:
Here:https://github.com/Morganamilo/paru/blob/eef68032b399c6d73a7247af20f73602042280b8/src/download.rs#L317

jimman2003:
error e0716...Should I fix it by removing the borrow?

Morganamilo:
If that fixes it sure. I'm not aware of what how the new version differs.

jimman2003:
Done!

Morganamilo:
It looks like you're doing this all by one. You should install cargo-edit and use that.

jimman2003:
Thanks for the advice..

jimman2003:
OK..cargo-edit confirmed the deps are up-to-date

PR #423, opened 2021-06-25
Overhaul release CI jobs to automate changelog

alerque:
Given your recent [AUR comment](https://aur.archlinux.org/packages/paru/#comment-814472):

> The CI makes a draft release. I hate automated releases and it lets me write the changelog in before publishing.

I couldn't resist. Marking automated releases as drafts until hand checked and everything looks copacetic I can understand. Writing release notes by hand when you have a changelog file anyway I cannot.

The actual changelog is relatively simple:

```console
$ sed -ne '/v1.7.3/,/^## /{/^## /d;p}' CHANGELOG.md | sed -e '1d;$d'
```

As a post-script, allow me to note that your workflows look overly convoluted to me. I'm pretty sure we could get the same job done in a much less round-about way. The actions you are using for this are clumsy and not even supported any more (at least one of them is completely deprecated by GitHub officially). Would you be open to accepting an overhaul of the entire workflow process to simplify all the value passing / stages / artifact push & pulls / etc.?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,7 @@ name: Build Release
 on:
   push:
     tags:
-      - v*
+      - v[0-9]+.[0-9]+.[0-9]+
 
 jobs:
   build-releases:
@@ -59,6 +59,8 @@ jobs:
     needs: [build-releases]
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
       - name: Read info
         id: tags
         shell: bash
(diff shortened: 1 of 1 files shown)


Morganamilo:
I hate the CI and hate github actions so go ahead. 

The change to the tag format though is wrong and I don't know why you've changed it.

Also the sed is wrong and doesn't include the changelog or version header.

Morganamilo:
Also if you can get 32 bit builds working that would be nice too :P

jclds139:
I updated the existing workflow process to get 32-bit arm builds working in #723, so that can at least give you a reference to include them in your overhaul.

PR #428, opened 2021-06-27
Localize paru

Morganamilo:
See CONTRIBUTING.md for instructions


src/config.rs
```diff
@@ -730,7 +737,7 @@ impl Config {
         if key == "Include" {
             let value = match value {
                 Some(value) => value,
-                None => bail!("value can not be empty for value '{}'", key),
+                None => bail!(tr!("value can not be empty for value '{}'", key)),
```
ouuan, line 740:
```suggestion
                None => bail!(tr!("value can not be empty for key '{}'", key)),
```

src/config.rs
```diff
@@ -860,7 +870,7 @@ impl Config {
         let has_value = value.is_some();
         let value = value
             .map(|s| s.to_string())
-            .ok_or_else(|| anyhow!("value can not be empty for value '{}'", key));
+            .ok_or_else(|| anyhow!(tr!("value can not be empty for value '{}'", key)));
```
ouuan, line 873:
```suggestion
            .ok_or_else(|| anyhow!(tr!("value can not be empty for key '{}'", key)));
```

src/config.rs
```diff
@@ -746,20 +753,20 @@ impl Config {
...
             Some(section) => section.as_str(),
-            None => bail!("key '{}' does not belong to a section", key),
+            None => bail!(tr!("key '{}' does not belong to a section", key)),
         };
 
         match section {
             "options" => self.parse_option(key, value),
             "bin" => self.parse_bin(key, value),
-            _ => bail!("unknown section '{}', section"),
+            _ => bail!(tr!("unknown section '{}', section")),
```
ouuan, line 762: This is a little bit confusing. ", section" shouldn't be translated, right? I think it shouldn't be wrapped in `tr()`.

ouuan, line 762:
```suggestion
            _ => bail!(tr!("unknown section '{}'"), section),
```

ouuan:
When there are multiple `{}`s in the string, it's impossible to change the order, and in some languages, it will be strange if we try to keep the order of `{}`s :thinking:

In Qt, `%1`, `%2`, etc. is used instead of `{}` so that it's easy to change the order. Is there something like this in Rust?

src/devel.rs
```diff
@@ -122,7 +123,11 @@ pub async fn gendb(config: &mut Config) -> Result<()> {
 
         !devel_info.info.contains_key(pkg)
     });
-    println!("{} {}", action.paint("::"), bold.paint("Querying AUR..."));
+    println!(
+        "{} {}",
+        action.paint("::"),
+        bold.paint(tr!("Querying AUR...)"))
```
ouuan, line 129: What's this closing bracket? Is this expected?
```suggestion
        bold.paint(tr!("Querying AUR..."))
```

Morganamilo:
> When there are multiple `{}`s in the string, it's impossible to change the order, and in some languages, it will be strange if we try to keep the order of `{}`s 🤔
> 
> In Qt, `%1`, `%2`, etc. is used instead of `{}` so that it's easy to change the order. Is there something like this in Rust?

Yes you can do '{:foo}` that needs to be changed in the code though. I'll get to it.

src/devel.rs
Morganamilo: Probably just an error from refactoring.

src/download.rs
```diff
@@ -222,9 +228,16 @@ fn repo_pkgbuilds<'a>(config: &Config, pkgs: &[Targ<'a>]) -> Result<i32> {
     }
 
     if !missing.is_empty() {
-        let len = ":: Missing ABS packages ".len();
-        print!("{} Missing ABS packages ", config.color.error.paint("::"));
-        print_indent(config.color.base, len, 3, config.cols, "  ", &missing);
+        let msg = tr!("Missing ABS packages ");
+        print!("{}", tr!("{} {} ", msg, config.color.error.paint("::")));
```
ouuan, line 232:
```suggestion
        print!("{} {} ", config.color.error.paint("::"), msg);
```

src/util.rs
```diff
@@ -75,7 +76,11 @@ pub fn split_repo_aur_targets<'a, T: AsTarg>(
 pub fn ask(config: &Config, question: &str, default: bool) -> bool {
     let action = config.color.action;
     let bold = config.color.bold;
-    let yn = if default { "[Y/n]:" } else { "[y/N]:" };
+    let yn = if default {
+        tr!("[Y/n]:")
+    } else {
+        tr!("[y/N]:")
+    };
```
ouuan, line 83: I think this shouldn't be translated?

src/config.rs
Morganamilo: Looks like the tr macro doesn't panic if there's the amount of `{}` and values don't match up. That's unfortunate.

Morganamilo:
@ouuan  Thanks for looking over, Should have taken care of everything now.

PR #430, opened 2021-06-28
fix(src/clean.rs): -Sc not clean untracked files

akiirui:
Because `keep_installed` and `keep_current` return `Ok(())`, `clean_untracked()` never runs.

So, the actual result is different from that in the manpage.
> -Sc    Paru will also clean cached AUR packages and any untracked files in the cache.

### DIFF
--- a/src/clean.rs
+++ b/src/clean.rs
@@ -131,6 +131,8 @@ fn clean_aur_pkg(
 
     let srcinfo = Srcinfo::parse_file(file.path().join(".SRCINFO"))?;
 
+    clean_untracked(config, &config.fetch.clone_dir.join(&srcinfo.base.pkgbase))?;
+
     if keep_installed {
         let local_db = config.alpm.localdb();
         for pkg in &srcinfo.pkgs {
@@ -153,7 +155,7 @@ fn clean_aur_pkg(
         }
     }
 
-    clean_untracked(config, &config.fetch.clone_dir.join(srcinfo.base.pkgbase))
+    Ok(())
 }
 
 pub fn clean_untracked(config: &Config, path: &Path) -> Result<()> {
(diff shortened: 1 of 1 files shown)


Morganamilo:
`-Sc` respects `CleanMethod =`. So `-Sc` won't act on anything it's set not to `-Scc` is to act on everything.

I guess it's not explicitly mentioned in the man page because paru inherits from pacman so tries not to repeat things from pacman's man page.

akiirui:
> `-Sc` respects `CleanMethod =`. So `-Sc` won't act on anything it's set not to `-Scc` is to act on everything.

Yep, but `clean_untracked()` never runs from any methods. Because these lines `return Ok(())`

https://github.com/Morganamilo/paru/blob/f691bc283bcfa4ecbcd955ea4944db7b22cbd558/src/clean.rs#L134-L142

https://github.com/Morganamilo/paru/blob/f691bc283bcfa4ecbcd955ea4944db7b22cbd558/src/clean.rs#L145-L154

And this line never runs:

https://github.com/Morganamilo/paru/blob/f691bc283bcfa4ecbcd955ea4944db7b22cbd558/src/clean.rs#L156

If you think `clean_untracked()` should only runs with `-Scc (keep_current)`, we need move `clean_untracked()` into https://github.com/Morganamilo/paru/blob/8de16d6d30c21528a2af7b9366636cac0fdf8a06/src/clean.rs#L147-L156

Morganamilo:
Ah right. The check may have accidentally deleted during refactoring.

In that case the two ifs should be wrapped in `if config.clean == 1 {}`.

akiirui:
Oh, bad. I accidentally deleted the previous comment.

akiirui:
> Ah right. The check may have accidentally deleted during refactoring.
> 
> In that case the two ifs should be wrapped in `if config.clean == 1 {}`.

Yep, follow your way.

`-Sc` not clean the untracked files, and the `-Scc` will clean the untracked files.

Morganamilo:
Looks good to me. This may be the reason for some of the random -Sc issues.

PR #436, opened 2021-06-28
Added de_DE localisation

mocdaniel:
I tried to translate to the shortest possible idiomatic German equivalents, some got rather long anyways, so maybe somebody would like to review this.

Does paru support localised input for [Y/n] dialogues and the likes? When translating, I pretended it did, this might need correction depending on paru's behaviour (e.g. see [J/n] translation).

Some rather technical terms such as `make`, `build`, `repository` and `maintainer` remain the same in German.


Morganamilo:
Yes pacman localized y/n so paru should do the samep can you make this just he instead of de_DE though to copy pacman?

po/de_DE.po
```diff
@@ -0,0 +1,706 @@
...
+msgid "unknown option -{}"
+msgstr "Unbekannte Option -{}"
+
+#: src/completion.rs:27
+msgid "failed to open cache file '{}'"
+msgstr "Cache-Datei ‚{}‘ konnte nicht geöffnet werden"
+
+#: src/completion.rs:65
+msgid "could not update aur cache"
+msgstr "AUR-Cache konnte nicht geupdated werden"
```
daniel-seiler, line 78:
```suggestion
msgstr "AUR-Cache konnte nicht aktualisiert werden"
```

po/de_DE.po
```diff
@@ -0,0 +1,706 @@
...
+msgstr "Option ‚{}‘ nimmt keinen Wert"
+
+#: src/config.rs:947 src/install.rs:1796
+msgid "There are {:n} providers available for {:pkg}:"
+msgstr "Es sind {:n} Anbieter für {:pkg} verfügbar:"
+
+#: src/config.rs:961 src/info.rs:56 src/info.rs:95 src/install.rs:1755
+#: src/install.rs:1804
+msgid "Repository"
+msgstr "Repository"
```
daniel-seiler, line 143: In the [pacman repo,](https://gitlab.archlinux.org/pacman/pacman/-/blob/master/src/pacman/po/de.po) "Repository" translates to "Repositorium". Sounds unnatural, but might be worth a thought.

po/de_DE.po
```diff
@@ -0,0 +1,706 @@
...
+msgid "failed to get current exe"
+msgstr "Aktuelle exe konnte nicht gefunden werden"
+
+#: src/repo.rs:195
+msgid "syncing local databases..."
+msgstr "Synchronisiere lokale Datenbanken…"
+
+#: src/repo.rs:201
+msgid "  nothing to do"
+msgstr "Nichts zu tun"
```
daniel-seiler, line 586: Because of the spaces in front, the string might be used as continuation of a sentence. In this case:
```suggestion
msgstr " nichts zu tun"
```
would be appropriate.

daniel-seiler (commented):
To sum up:
- IMO the technical terms are understandable (unless the r/ich_iel approach is desired)
- For some strings it isn't clear if they're used as a continuation of a sentence. If that's the case, some German strings need to start lower case.
Otherwise I liked the translation.

Morganamilo:
Looking through you seem to be Capitalising everything. Please Keep the case of the original message.

mocdaniel:
I added the suggestions by @daniel-seiler and reduced capitalisation where possible. Due to German syntax, many of the translations start with nouns which are always capitalised in German. I also changed the filename, to `de.po`.

Morganamilo:
> Due to German syntax, many of the translations start with nouns which are always capitalised in German.

Oh that's fine in that case.

po/de_DE.po
Morganamilo: This still needs to be fixed.

mocdaniel: Done, the double whitespace slipped my attention.

PR #440, opened 2021-06-28
Initial upload for Italian translation

BachoSeven:
The only opinionated choice I think I made is to not translate `directory`, since I noticed this is also the case for the Italian translation of `pacman`.

A review would be appreciated if anyone wants to do that

### DIFF
--- a/po/it.po
+++ b/po/it.po
@@ -1,0 +1,704 @@
+# Translation of paru into Italian.
+# This file is distributed under the same license as the paru package.
+# Francesco Minnocci <ascoli@minnocci>, 2021.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: paru 1.7.3\n"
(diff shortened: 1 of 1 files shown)


po/it.po
```diff
@@ -0,0 +1,704 @@
...
+msgid "can not install conflicting packages with --noconfirm"
+msgstr "impossibile installare pacchetti in conflitto con --noconfirm "
+
+#: src/install.rs:1117
+msgid "Repo"
+msgstr "Repo"
+
+#: src/install.rs:1124
+msgid "Repo Make"
+msgstr "Repo (compilazione)"
```
Morganamilo, line 494: I actually like this a lot. May switch the English to `Repo (make):`

Morganamilo:
Looks good to me on the formatting front. Thanks.

PR #441, opened 2021-06-28
Finnish translation for paru

keinhen:
Simple translation, some technical terminology (make / devel / snapshot) was left unchanged for clarity over full translation, am open to suggestions if there are any commonly used terms for these!

Tried to follow pacman's own finnish translation for reference on other technical terminology such as repository (pakettivarasto), but in some parts such as "repo package", tried to avoid redundancy of "pakettivarasto paketti" with simpler, "varasto paketti", but am not sure if this should just be replaced by "repo paketti" or something more simple and similar.

### DIFF
--- a/po/fi.po
+++ b/po/fi.po
@@ -1,0 +1,688 @@
+# Paru Finnish translation 
+# Copyright (C) YEAR THE paru'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the paru package.
+# Henry Keinälä <henry.keinala@tuta.io> 2021
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: paru VERSION\n"
+"Report-Msgid-Bugs-To: https://github.com/Morganamilo/paru\n"
+"POT-Creation-Date: 2021-06-28 03:24+0000\n"
+"PO-Revision-Date: 2021-06-28 10:08+0200\n"
+"Last-Translator: Henry Keinälä <henry.keinala@tuta.io>\n"
+"Language-Team: \n"
+"Language: fi\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+
+#: src/clean.rs:31
+msgid "Do you want to remove ALL AUR packages from cache?"
+msgstr "Halutako poistaa KAIKKI AUR paketit välimuistista?"
+
+#: src/clean.rs:33
+msgid "Do you want to remove all other AUR packages from cache?"
+msgstr "Haluatko poistaa kaikki muut AUR paketit välimuistista?"
+
+#: src/clean.rs:40
+msgid "Clone Directory: {}"
+msgstr "Kloonaa kansio: {}"
+
+#: src/clean.rs:46
+msgid "\n"
+"Diff Directory: {}"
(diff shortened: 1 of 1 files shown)


po/fi.po
```diff
@@ -0,0 +1,688 @@
...
+msgid "yes"
+msgstr "kyllä"
+
+#: src/util.rs:103
+msgid "n"
+msgstr "e"
+
+#: src/util.rs:103
+msgid "no"
+msgstr "Ei"
```
Morganamilo, line 668: Case needs to be maintained.

po/fi.po
```diff
@@ -0,0 +1,688 @@
...
+
+#: src/install.rs:762
+msgid "failed to read dir: {}"
+msgstr "ei pystytty lukemaan kansiota: {}"
+
+#: src/install.rs:778
+msgid "{} is a directory\n"
+"{} on kansio\n"
+""
+msgstr ""
```
Morganamilo, line 413: Not translated.

po/fi.po
```diff
@@ -0,0 +1,688 @@
...
+msgstr "duplikaatti paketti: {}"
+
+#: src/install.rs:975
+msgid "could not find all required packages:"
+msgstr "ei pystytty löytämään kaikkia vaadittuja paketteja"
+
+#: src/install.rs:982
+msgid "\n"
+"    {:missing} (wanted by: {:stack})"
+msgstr ""
```
Morganamilo, line 446: Not translated.

Morganamilo:
looks good to me at a glance apart from the couple comments.

po/fi.po
keinhen: Fixed this in the latest commit

po/fi.po
```diff
@@ -408,9 +408,9 @@ msgstr "ei pystytty lukemaan kansiota: {}"
 
 #: src/install.rs:778
 msgid "{} is a directory\n"
-"{} on kansio\n"
+"\n"
 ""
-msgstr ""
+msgstr "{} on kansio\n"
```
Morganamilo, line 413: You need to follow the exact formatting as above.

po/fi.po
keinhen: Fixed this in the latest commit. 

I gather the {:text} are variables to be replaced by package names? 

In this context I wasn't able to add in a satisfactory, simple translation so I had left it empty, but this time commit I offered in at least a placeholder.

po/fi.po
keinhen: My bad, sorry. Should be alright in the latest with identical formatting to msgid""

po/fi.po
Morganamilo: Yes though I realise it's not working so I need to look into that hmmm

po/fi.po
Morganamilo: Yes. I only commented on the first occurrence too. You need to do this elsewhere.

keinhen: Ah gotcha. I gave it a once over for formatting rigor.

Morganamilo:
Looks good now from what I can see. Though I am getting a little worn out with all these :P Thanks

keinhen:
> Looks good now from what I can see. Though I am getting a little worn out with all these :P Thanks

Great! And no problem, can imagine it can be hard to keep the focus up after a few rounds :D

PR #442, opened 2021-06-28
portuguese (pt-pt) translation small issue fix: change error to problema

jcfb-opensource:
sorry for making another pull request but i forgot to change some lines


Morganamilo:
It's fine. I don't care if some errors leak in when I merge the initial translation. Makes the follow up fixes really easy because the diffs should be small.

PR #447, opened 2021-06-29
Add pl_PL translations

rail01:
pl.po file containing Polish translations for paru. If strings are missing, it means that I was not sure about the context and decided not to risk incorrect translation.

Btw this program has very straightforward attitude towards plurals. This will become an issue for languages with more complex plurals, such as Polish. Just compare these two:

https://gitlab.archlinux.org/pacman/pacman/-/blob/master/src/pacman/po/pl.po#L1511
```
#: src/pacman/sync.c:579
#, c-format
msgid "There is %d member in group %s%s%s:\n"
msgid_plural "There are %d members in group %s%s%s:\n"
msgstr[0] "Znaleziono %d pakiet w grupie %s%s%s:\n"
msgstr[1] "Znaleziono %d pakiety w grupie %s%s%s:\n"
msgstr[2] "Znaleziono %d pakietów w grupie %s%s%s:\n"
msgstr[3] "Znaleziono %d pakietów w grupie %s%s%s:\n"
```

And paru's:
```
#: src/install.rs:1743
msgid "There are {} members in group"
msgstr "Znaleziono {} pakietów w grupie"
```

### DIFF
--- a/po/pl.po
+++ b/po/pl.po
@@ -1,0 +1,694 @@
+# paru Polish translation
+# Copyright (C) 2021 THE paru'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the paru package.
+# Rail <rail01@protonmail.com> 2021
+# Pacman Polish translators
+#
+#, fuzzy
+msgid ""
+msgstr ""
+"Project-Id-Version: paru VERSION\n"
+"Report-Msgid-Bugs-To: https://github.com/Morganamilo/paru\n"
+"POT-Creation-Date: 2021-06-28 03:24+0000\n"
+"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+"Language-Team: LANGUAGE <LL@li.org>\n"
+"Language: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+
+#: src/clean.rs:31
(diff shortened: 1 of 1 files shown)


Morganamilo:
How does the singular and plural become 4 translations. I'm assuming singular/plural*masc/fem. How does the code actually how which one to translate to?

rail01:
I would love to tell you, but I don't know :) I'm more of a translatior than a developer – sorry.

trevorgunn:
You mean you have a different plural for 0 items, 1 item, 2 items, 3 or more items right? Not male/female.

Morganamilo:
That would make more sense. I assume `msgstr[3]` is used if count is > 3? Need to look into gettext.

Morganamilo:
Ah, it's dependant on "Plural-Forms" at the top of the file. So the polish translation defines 4 plural forms and that maps to msgstr[n].

Morganamilo:
Anyway I'll worry about that in a future release when all the initial translations are in place.

If you can finish the translation then please do. Otherwise I'll leave this open in case some one else can fill in the gaps.

Cyliann:
Polish is very complicated in therms of numerals. The are 3 plural forms for every gramatical gender (male/female/neuter). 
Pakiet (package) is male type so:
1 package -> pakiet
2-4 packages -> pakiety
5+ packages -> pakietów

That's not all. If a number ends with 2, 3 or 4 it's still pakiety. Eg. 23 pakiety, but 21 pakietów.

mskiptr:
Just for completeness sake:

0 packages -> pakietów
> 1 package -> pakiet
> 2-4 packages -> pakiety
> 5+ packages -> pakietów

PR #449, opened 2021-06-29
Create a Russian translation

Fifis:
Initial upload, using the pre-existing Chinese PO as a template


Morganamilo:
Looks good to me. Though this is one of the languages that is the least readable to me.

PR #435, opened 2021-06-28
Add es localization

alosarjos:
Spanish (Spain) localization. 

PD: Thanks for the paru project. Always nice to see more Rusty stuff

PR #456, opened 2021-07-04
Add fr translation

Paragoumba:
Translated all the strings to french but some might need tweaking

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -1,15 +1,15 @@
-# SOME DESCRIPTIVE TITLE.
+# Paru French translation
 # Copyright (C) YEAR THE paru'S COPYRIGHT HOLDER
 # This file is distributed under the same license as the paru package.
-# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
+# Paragoumba <paragoumba@gmail.com>, 2021.
 #
 msgid ""
 msgstr ""
 "Project-Id-Version: paru VERSION\n"
 "Report-Msgid-Bugs-To: https://github.com/Morganamilo/paru\n"
 "POT-Creation-Date: 2021-07-02 13:11+0000\n"
-"PO-Revision-Date: 2021-07-02 15:00+0200\n"
-"Last-Translator: \n"
+"PO-Revision-Date: 2021-07-04 22:16+0200\n"
+"Last-Translator: Paragoumba <paragoumba@gmail.com>\n"
 "Language-Team: \n"
 "Language: fr\n"
 "MIME-Version: 1.0\n"
@@ -78,7 +78,6 @@ msgid "could not update aur cache"
 msgstr "impossible de mettre à jour le cache de l'AUR"
 
 #: src/config.rs:174
-#, fuzzy
 msgid "invalid value '{val}' for key '{key}', expected: {exp}"
 msgstr "valeur '{val}' invalide pour la clé '{key}', attendu : {exp}"
 
@@ -135,7 +134,6 @@ msgid "option '{}' does not take a value"
 msgstr "l'option '{}' n'accepte pas de valeur"
 
 #: src/config.rs:947 src/install.rs:1796
-#, fuzzy
 msgid "There are {n} providers available for {pkg}:"
 msgstr "Il y a {n} fournisseurs disponibles pour {pkg} :"
 
@@ -181,7 +179,6 @@ msgid "packages not in the AUR: "
 msgstr "paquets introuvables dans l'AUR : "
 
 #: src/download.rs:94
-#, fuzzy
 msgid "marked out of date: "
 msgstr "marqués comme pas à jour : "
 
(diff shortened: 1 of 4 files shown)


po/fr.po
```diff
@@ -0,0 +1,722 @@
...
+msgid "unknown option --{}"
+msgstr "option inconnue --{}"
+
+#: src/command_line.rs:321
+msgid "unknown option -{}"
+msgstr "option inconnue-{}"
+
+#: src/completion.rs:27
+msgid "failed to open cache file '{}'"
+msgstr "erreur lors de l'ouverture du fichier de cache '{}'"
```
Morganamilo, line 74: See https://github.com/Morganamilo/paru/pull/435#discussion_r660053767

po/fr.po
```diff
@@ -0,0 +1,722 @@
...
+
+#: src/completion.rs:65
+msgid "could not update aur cache"
+msgstr "impossible de mettre à jour le cache de l'AUR"
+
+#: src/config.rs:174
+#, fuzzy
+msgid "invalid value '{val}' for key '{key}', expected: {exp}"
+msgstr "valeur '{:val}' invalide pour la clé '{:key}', attendu: {:exp}"
```
Morganamilo, line 83: It's `{val}` not `{:val}`.

Paragoumba: Yes a PR has been opened on my fork. It will be fixed soon

po/fr.po
```diff
@@ -0,0 +1,706 @@
...
+msgid "packages failed to build: {}"
+msgstr "la compilation des paquets suivants a échouée : {}"
+
+#: src/install.rs:698
+msgid "could not find .SRINFO for '{}'"
+msgstr "impossible de trouver le .SRCINFO pour '{}'"
+
+#: src/install.rs:762
+msgid "failed to read dir: {}"
+msgstr "échec de la lecture du dossier : {}"
```
Morganamilo, line 416: is `foo : bar` intentional?

Paragoumba: The space before :, yes

Morganamilo: Then you're missing it in other places unless there#s some context i'm unaware of.

Paragoumba: I checked multiple times and didn't find a place that lacks a space before a colon

Morganamilo: Ah may have looked at the English by mistake.

po/fr.po
```diff
@@ -0,0 +1,706 @@
...
+#: src/download.rs:543 src/download.rs:557 src/exec.rs:72 src/exec.rs:113
+#: src/exec.rs:122 src/exec.rs:148 src/exec.rs:161 src/exec.rs:186
+#: src/exec.rs:200 src/exec.rs:221 src/exec.rs:234 src/install.rs:744
+#: src/install.rs:807 src/install.rs:845 src/keys.rs:74
+msgid "failed to run:"
+msgstr "échec du lancement :"
+
+#: src/download.rs:263
+msgid "does not contain PKGBUILD: skipping"
+msgstr "ne contient pas de PKGBUILD : ignorer"
```
michaelb, line 211: I wonder if the appropriate wording would be 'ignoré' rather than 'ignorer'. In a context where 'skipping' made more sense than 'skip', it looks better imo

po/fr.po
```diff
@@ -0,0 +1,706 @@
...
+msgid " ({total>padding$}/{total}) downloading: {pkg}"
+msgstr " ({total>padding$}/{total}) téléchargement : {pkg}"
+
+#: src/download.rs:317
+msgid "Downloading PKGBUILDs..."
+msgstr "Téléchargement des PKGBUILDs..."
+
+#: src/download.rs:321
+msgid " PKGBUILDs up to date"
+msgstr " Les PKGBUILDs sont à jour"
```
michaelb, line 223: 'PKGBUILDs à jour'

Wouldn't that be closer a translation ?

po/fr.po
michaelb: To be consistent with line 464 also

Silejonu: Line 464 should be changed to "ignorer", in my opinion. "ignoré" would be "skipped".

po/fr.po
Silejonu: > is `foo : bar` intentional?

Yes, spacing rules and conventions are different in French when it comes to punctuation.

po/fr.po
Paragoumba: "ignorer" is more literal but I think "ignoré" would be more understandable in this context as it could be read as "ne contient pas de PKGBUILD il a été ignoré"

po/fr.po
Paragoumba: Yes i will change that

po/fr.po
Paragoumba: I changed line 211 to "ignoré" to be more consistent with line 464. Tell me if you think that's a mistake

Silejonu: Neither "ignorer" nor "ignoré" sound like mistakes to me. I initially favoured "ignorer" to stick as close as possible to the original, but since this string isn't likely to be used in several different contexts, I think it's fine to be less literal in the translation, and gain in understandability.

Morganamilo: Look at what pacman uses in its translation.

Paragoumba: [The only "skipping" that appears in th english version has been translated to "omission"](https://gitlab.archlinux.org/pacman/pacman/-/blob/master/src/pacman/po/fr.po#L1549). As this translation works in this case, I think we're settled

PR #468, opened 2021-07-15
Fix various problems in the French translation

Paragoumba:


Morganamilo:
I'm told the space before colons is intentional. Not sure why you're taking spaces out of the help text. The missing \n's do look valid though.

Paragoumba:
Yes they it is. I wanted to remove it because it might look weird in a CLI. I've seen that pacman has it so it might be better to keep it for consistency

po/fr.po
```diff
@@ -312,40 +312,35 @@ msgstr "Si aucun argument n'est fourni 'paru -Syu' sera exécuté"
...
 
 #: src/help.rs:30
 msgid "New options:"
-msgstr "Nouvelles options:"
+msgstr "Nouvelles options :"
 
 #: src/help.rs:31
 msgid "       --repo              Assume targets are from the repositories"
-msgstr "       --repo              Suppose que les cibles sont dans les depôts"
+msgstr "       --repo             Suppose que les cibles sont dans les depôts"
```
Morganamilo, line 331: Why did you remove this? It pushes everything back and misalignes it.

Paragoumba: I removed the space so everything (except the flags from --makepkg to --fmflags included) is aligned. The <n> of --completioninterval is aligned with the other flags descriptions

Morganamilo: That makes there an only 1 space gap between `combinedupgrade` though where there should be at least two. The misalignment there probably happened when the `--[no]`s were added. But anyway just stick to the english layout and that can be fixed another time.

Paragoumba: I fixed the spaces

PR #476, opened 2021-07-22
Add openssl and rustls features

dodomorandi:
This could be the first step toward improving the situation around #83: if we can remove dynamic libraries, it is much easier to cross compile. Obviously, `openssl` feature is by default to avoid breaking changes.

Unfortunately, this is not enough to cross-compile against armv6l, it looks like `ring` is not happy.

This is blocked by https://github.com/Morganamilo/aur-depends/pull/4

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -100,8 +100,6 @@ dependencies = [
 [[package]]
 name = "aur-depends"
 version = "1.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e2284dabc5843dcf1e8ea31656351ac00070b16617bacf3a16529721cd0069f8"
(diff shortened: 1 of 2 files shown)


Morganamilo:
#83 is blocked by cargo erroring out weirdly on 32 bit arches. paru links to alpm which in turn links to openssl so I don't see any point in this being a feature.

dodomorandi:
I understand, and if you think that this is not worth, close the PR without worries. I noticed that you were already supporting both `openssl` and `rustls` in your dependencies, I thought that it could be a very tiny helpful step.

Morganamilo:
Yeah it's in the raur it makes a bit of sense because you're nor locked into openssl via alpm.

PR #482, opened 2021-07-30
update paru.8 & paru.conf.5 to suppress troff warning

peterwu:
When running `man paru > /dev/null` or `man paru.conf > /dev/null`, the system reports
```
troff: <standard input>:1: name expected (got '\ '): treated as missing
```
The culprit appears to be the first line in the paru.8 & paru.conf.5 files. This PR fixes this problem.

### DIFF
--- a/man/paru.8
+++ b/man/paru.8
@@ -1,4 +1,3 @@
-'\ t
 .TH "PARU" "8" "2020\-11\-2" "paru v1.0.2 "Paru Manual"
 .nh
 .ad l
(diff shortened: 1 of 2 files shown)


Morganamilo:
I did want to look why this was there originally and as far as I can tell it's copied from pacman's man page which should be part of a comment but a bit is missing making it not a comment. It's been there so long I have no way to tell if it's intentional or not so I guess it's best just gone instead of fixed.

PR #485, opened 2021-07-31
Fix '-r' and '-b' crashing paru

Kanjirito:
The short global pacman flags were not using the given value resulting in paru setting the respective paths to an empty string which caused a `error: failed to initialize alpm: root=/ dbpath=: could not find or read directory` crash.

### DIFF
--- a/src/command_line.rs
+++ b/src/command_line.rs
@@ -101,10 +101,10 @@ impl Config {
             let mut chars = arg.chars();
             chars.next().unwrap();
 
-            while let Some(c) = chars.next() {
+            for c in chars {
                 let arg = Arg::Short(c);
                 if takes_value(arg) == TakesValue::Required {
-                    self.handle_arg(arg, Some(chars.as_str()), op_count, false)?;
+                    self.handle_arg(arg, value, op_count, false)?;
                     return Ok(true);
                 }
                 self.handle_arg(arg, None, op_count, false)?;
--- a/src/command_line.rs
+++ b/src/command_line.rs
@@ -101,11 +101,16 @@ impl Config {
             let mut chars = arg.chars();
             chars.next().unwrap();
 
-            for c in chars {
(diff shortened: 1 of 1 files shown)


Morganamilo:
I don't really get what the problem is nor how this fixes it.

Morganamilo:
I think what you're actually looking for is:

```diff
diff --git a/src/command_line.rs b/src/command_line.rs
index 2030474..bb732e6 100644
--- a/src/command_line.rs
+++ b/src/command_line.rs
@@ -104,7 +104,11 @@ impl Config {
             while let Some(c) = chars.next() {
                 let arg = Arg::Short(c);
                 if takes_value(arg) == TakesValue::Required {
-                    self.handle_arg(arg, Some(chars.as_str()), op_count, false)?;
+                    if chars.as_str().is_empty() {
+                        self.handle_arg(arg, value, op_count, false)?;
+                    } else {
+                        self.handle_arg(arg, Some(chars.as_str()), op_count, false)?;
+                    }
                     return Ok(true);
                 }
                 self.handle_arg(arg, None, op_count, false)?;
```

Kanjirito:
The problem is that `paru -b /tmp/db` doesn't work but `pacman -b /tmp/db` works. Same for `-r`. The `/tmp/db` gets ignored and paru sets the database path to an empty string which obviously won't work. I wasn't aware that `pacman -b/tmp/db` is valid. Doing it the way you posted will make it work with the space or without it but it will skip the next argument when used without the space. So `paru -b/tmp/db -Ss neofetch` basically runs `paru -b/tmp/db neofetch`. This is also true for the current release but not for pacman. Returning `Ok(true)` in the first if case and `Ok(false)` in the second one seems to solve this.

Morganamilo:
Yeah that makes sense. Funny it would appear I've never actually used the `-r foo` form to notice it's broken. PR looks good now thanks.

PR #487, opened 2021-08-05
use active form for german strings, instead of passive

soloturn:


Morganamilo:
@mocdaniel

po/de.po
```diff
@@ -643,19 +643,19 @@ msgstr " [installiert]"
...
+msgstr "Suche devel-Upgrades"
 
 #: src/upgrade.rs:179
 msgid "warning:"
 msgstr "Warnung:"
 
 #: src/upgrade.rs:181
 msgid "{pkg}: ignoring package upgrade ({old} => {new})"
-msgstr "{pkg}: Ignoriere Paket-Upgrade ({old} => {new})"
+msgstr "{pkg}: Ignoriere aktualisieren von ({old} => {new})"
```
mocdaniel, line 658: why go away from the original wording?

po/de.po
```diff
@@ -336,27 +336,27 @@ msgstr "keine Architektur"
 
 #: src/install.rs:138
 msgid "can't build package as root"
-msgstr "Pakete können nicht als root gebaut werden"
+msgstr "kann Pakete nicht als root bauen"
 
 #: src/install.rs:142
 msgid "failed to parse srcinfo generated by makepkg"
-msgstr "von Makepkg generierte srcinfo konnte nicht geparsed werden"
+msgstr "kann von Makepkg generiertes srcinfo nicht parsen"
```
mocdaniel, line 343: maybe `generierte` instead of `generiertes`  since `info` takes female gender in German

po/de.po
```diff
@@ -144,31 +144,31 @@ msgstr "Repositorium"
...
 
 #: src/devel.rs:193
 msgid "failed to save devel info"
-msgstr "devel Informationen konnten nicht gespeichert werden"
+msgstr "kann devel Informationen nicht speichern"
 
 #: src/devel.rs:382
 msgid "package does not have an update"
-msgstr "für das Paket gibt es kein Update"
+msgstr "Paket ist in der neuesten Version"
```
mocdaniel, line 167: Imo this is wrong. "Package X is the most current version" has other implications than "Package X does not have an update", especially for AUR packages.

mocdaniel (changes requested):
I added some comments. Overall, the changes are fine and indeed shorter and closer to the English formulations, some points should be changed in my opinion, however.

po/de.po
soloturn: oh, i see, this is a little unfortunate. das file, die datei, die info. in doubt i tend to use neutral, instead of female or male ...

soloturn: reverted all others you pointed out @mocdaniel - thanks!

soloturn:
reverted all but one @mocdaniel , thanks!!

PR #492, opened 2021-08-06
Improve error message for failing makepkg command

mqus:
related to #491.

callers of `makepkg_output` never actually provide `--verifysource -Ccf`, but rather `--printsrcinfo` or `--packagelist`

Print the actual `args` instead.


Morganamilo:
This must have been copy pasted when refactored...

Anyway thanks.

PR #503, opened 2021-08-14
[Minor] Make upgrade text bold (to match pacman)

MithicSpirit:
When running commands like `pacman -Qu`, pacman prints the upgrade text in green and bold, while paru currently only prints it in green. This very very minor commit fixes that disparity.

image
image


Morganamilo:
Thanks never noticed.

PR #504, opened 2021-08-15
Update Japanese translation

livewing:
- Add Japanese translations for `src/help.rs`, etc.
- Change some existing translations because it was different from [pacman's](https://gitlab.archlinux.org/pacman/pacman/-/blob/master/src/pacman/po/ja.po).


Morganamilo:
@is-hoku

is-hoku:
It would be better.

Morganamilo:
You mean this PR is an improvement right? And not "this pr could be better"?

is-hoku:
Yes, I think this PR is fine. The Japanese Translation is more natural than before.

Issue #231, opened 2021-02-09
[Request] Don't put devel.json to .cache/paru


monosans:
I think that `devel.json` shouldn't be placed in `~/.cache` folder, because system cleaners (BleachBit, for example) delete all the content of `~/.cache` folder. It leads to paru forgetting which commit of package is installed.
I suggest putting `devel.json` in `XDG_STATE_HOME` or `XDG_DATA_HOME`.

soloturn:
but it is not really "shared" isn't it?

mqus:
`.local/share` is a default directory for user data[1]. Most of its content probably can't be shared, but there is no alternative default location. It's probably best to ignore the name

[1] https://developer.gnome.org/basedir-spec/
> $XDG_DATA_HOME defines the base directory relative to which user specific data files should be stored. If $XDG_DATA_HOME is either not set or empty, a default equal to $HOME/.local/share should be used.

monosans:
Fixed in https://github.com/Morganamilo/paru/pull/822.

PR #508, opened 2021-08-16
changed devel.json location to user data

meguminloli:
As @monosans pointed out in #231 , paru should store `devel.json` in `$XDG_DATA_HOME`, not in `$XDG_CACHE_HOME`, because some cleaners delete the cache. To keep compatibility, `devel.json` will be moved from `$XDG_CACHE_HOME` to `$XDG_DATA_HOME`.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -476,13 +476,27 @@ impl Config {
         let cache =
             dirs::cache_dir().ok_or_else(|| anyhow!(tr!("failed to find cache directory")))?;
         let cache = cache.join("paru");
+        
         let config =
             dirs::config_dir().ok_or_else(|| anyhow!(tr!("failed to find config directory")))?;
         let config = config.join("paru");
+        
+        let data = 
+            dirs::data_dir().ok_or_else(|| anyhow!(tr!("failed to find data directory")))?;
+        let data = data.join("paru");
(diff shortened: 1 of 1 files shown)


Morganamilo:
I'd rather it be in STATE_HOME but I don't believe dirs supports it.

meguminloli:
I took a look and dirs doesn't support STATE_HOME yet. I will open an issue there.

meguminloli:
Looks like only dirs-rs allows issues and pull requests only from past contributors.

Morganamilo:
Have you got the right repo?

meguminloli:
Isn't this one [dirs-rs](https://github.com/dirs-dev/dirs-rs)?

Morganamilo:
https://github.com/xdg-rs/dirs

meguminloli:
My mistake, didn't realize that it got changed. It looks like there is a pull request from 2020, that got blocked. https://github.com/xdg-rs/dirs/pull/43 . I will open an issue to remind them that STATE_HOME got adopted

ivalarian:
dirs 4.0.0 includes support for `$XDG_STATE_HOME` which we're already using, so this isn't blocked on that anymore.

PR #526, opened 2021-09-07
Fixed capitalization on 'marked out of date' to match convention

ericthomasca:
Having this line capitalised will make it match the other lines stating with two colons " : : "


Morganamilo:
In yay this was a submenu with a prefix of `->`. Looks like I've ditched the whole submenu thing and we never use `->` and only use `::`. In that cause it probably makes sense to capitalise it.

Morganamilo:
Looking through, key import and "does not contain PKGBUILD" should be capped too.

PR #536, opened 2021-10-03
Add -x/--regex for -Ss

Morganamilo:
Allows search the aur using regex

paru -Ssx ^p.ru


Geweldig:
I think this PR introduced a regression. The check to see if paru should use regex is
```if config.args.has_arg("s", "search")```
I believe this should depend on the `x` switch instead.

Morganamilo:
Looks like it. Copy paste mistake -_-

PR #537, opened 2021-10-04
Fix regex search condition

Geweldig:
Resolves regression introduced in #536, where regex search was always used for the aur.


Morganamilo:
Sorry I was a bit too fast and fixed it to moment you commented.

Geweldig:
Damn. At least it's fixed, that's the important part :smile:

Morganamilo:
Yeah thanks for noticing though.

PR #539, opened 2021-10-04
Add PreBuildCommand

Morganamilo:
This is a command run for each pkgbuild when installing packages

### DIFF
--- a/man/paru.conf.5
+++ b/man/paru.conf.5
@@ -293,6 +293,15 @@ being passed to file manager.
 .B Pager = Command
 Command to use for paging
 
+.TP
+.B PreBuildCommand = Command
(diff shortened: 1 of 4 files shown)


johnhamelink:
Just wanted to thank you for adding this, I switched to Paru from yay specifically because of this feature.

I'm using emacs 29 from the [emacs-git](https://aur.archlinux.org/packages/emacs-git) repo, and I want to stick to master while toggling options in the PKGBUILD periodically (such as enabling webkit browser support). I also don't want have to manually track the commit log, and strategically wait for a good moment to upgrade: I'd rather let the package maintainer make that call. Adding this function lets me use customizepkg, which lets me apply my own patches to the PKGBUILD before the package is updated.

Thank you!

PR #553, opened 2021-10-17
Update portuguese translation

hugok79:


po/pt.po
```diff
@@ -235,103 +245,461 @@ msgstr "asp devolveu {}"
...
+#: src/help.rs:79
+msgid ""
+"    --topdown              Shows repository's packages first and then AUR's"
+msgstr ""
+"    --topdown              Mostra primeiro os pacotes do repositório e "
+"depois os do AUR"
+
+#: src/help.rs:81
+msgid "show specific options:"
+msgstr "mostrar opções específicas:"
```
Morganamilo, line 576: show is the name of the arg and not just a word.

hugok79: @Morganamilo What do you mean? It shouldn't be translated, is that it?

po/pt.po
```diff
@@ -698,10 +1067,10 @@ msgstr "número inválido: {}"
 msgid "invalid value: {n} is not between 1 and {max}"
 msgstr "valor inválido: {n} não está entre 1 e {max}"
 
-#: src/lib.rs:153
+#: src/lib.rs:158
 msgid "can not use chroot builds: devtools is not installed"
-msgstr "não é possível compilar em chroot: devtools não está instalado"
+msgstr "não é possível compilar em chroot: o devtools não está instalado"
```
jcfb-opensource, line 1072: I think this is better without the 'o' like it was before

po/pt.po
```diff
@@ -648,7 +1017,7 @@ msgstr "A procurar atualizações no AUR"
 
 #: src/upgrade.rs:148
 msgid "Looking for devel upgrades"
-msgstr "A procurar atualizações de pacotes git"
+msgstr "À procura de atualizações de desenvolvimento"
```
jcfb-opensource, line 1020: I think "A procurar" is better than "à procura", most portuguese translations have the verb in infinitive form. In the pacman translation it is also "A procurar"

jcfb-opensource:
Check my review. Translation-wise it looks good to me. It definitely is better than my old translation.

po/pt.po
hugok79: Correction done

PR #562, opened 2021-10-25
Fix Clippy warnings, and upgrade dependencies.

ReperakDev:
There were no issues caused by upgrading and the test suite passes.

### Summary
* Clippy warnings fixed
* nix, rss upgraded (0.22.1 to 0.23.0 and 1.10.0 to 2.0.0, respectively)


Morganamilo:
Why is this closed?

PR #601, opened 2021-11-25
make paru -G return a correct exit code

### LINKED ISSUE #593: paru -G returns EXIT_SUCCESS even when it fails
At the moment, if `paru -G` fails, then the exit code is still zero (success). It would be better, if instead it returned a non-zero exit code (like e.g. `yay -G` does).

For example `paru -G asdf-not-existing-package` returns with 0, giving the impression that it was successful, but it would be better if it returned with a non-zero exit code.

This is also not in alignment with `paru -S asdf-not-existing-package`, which already does return error code 1.

getchoo:
Fixes #593

### DIFF
--- a/src/download.rs
+++ b/src/download.rs
@@ -181,7 +181,7 @@ pub async fn getpkgbuilds(config: &mut Config) -> Result<i32> {
         let warnings =
             cache_info_with_warnings(&config.raur, &mut config.cache, &aur, &config.ignore).await?;
         if !warnings.missing.is_empty() {
-            ret |= ret
+            ret = 1
         }
         warnings.missing(config.color, config.cols);
         let aur = warnings.pkgs;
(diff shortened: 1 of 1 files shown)


Morganamilo:
It should have been `ret |= !warnings.missing.is_empty() as i32` Not sure how it got mixed up.

getchoo:
Since it's in an if statement, wouldn't it make sense to just use `ret = 1` when `warnings.missing.is_empty()` is already evaluated so there wouldn't have to be another call to the `is_empty()` method? Other functions like `repo_pkgbuilds()` use this to set return values already.

Morganamilo:
Well that would be instead of the if statment. both ways work and some how the code some how has half of each.

vimpostor:
@getchoo I agree, replacing the whole if condition with `ret |= !warnings.missing.is_empty() as i32` would make more sense. If you replace the whole if condition instead of only the one line, `is_empty()` will not be evaluated twice.

getchoo:
@Morganamilo I see what you mean. Pushing another commit now.

Morganamilo:
To nitpick I don't see a need for the spaces around the line. Other than that looks good.

PR #603, opened 2021-11-30
allow more options to be configured in paru.conf and command line arguments

getchoo:
This pull request allows for more internal configuration options (namely `config.comments`, `config.makepkg_conf`, and `config.pager`) to be changed in the configuration file and passed as command line arguments. I updated the shell completions and the defualt `paru.conf` to reflect the changes as well.

I only added cli arguments that I found to be a configuration option in `src/config.rs` but not in `src/command_line.rs`. I did something similar with the configuration file options by adding configuration options that were avalible but not present in the `parse_*()` functions of `src/config.rs`. I checked a few times to make sure I got them all, ignoring some options I believed shouldn't be included

...but I wouldn't be surprised if I missed some

### DIFF
--- a/completions/bash
+++ b/completions/bash
@@ -70,7 +70,7 @@ _paru() {
            noconfirm noprogressbar noscriptlet quiet root verbose
 
            repo aur aururl clonedir makepkg mflags pacman git gitflags sudo sudoflags
-           asp gpg gpgflags fm fmflags completioninterval sortby searchby limit upgrademenu
+           asp gpg gpgflags fm fmflags pager completioninterval sortby searchby limit upgrademenu
            removemake noremovemake cleanafter nocleanafter rebuild rebuildall norebuild
            rebuildtree redownload noredownload redownloadall pgpfetch nopgpfetch useask
            nouseask savechanges nosavechanges combinedupgrade nocombinedupgrade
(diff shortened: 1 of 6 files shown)


Morganamilo:
The goal of the provided config is to be an example and highlight some options that may be useful. Not everything is meant to be listed there.

Comments is a flag and doesn't really make sense as a config option.

makepkgconf is good to add though. It also appears to be missing from the man page.

Pager is probably fine though. It may be useful for if paru is called from some non shell script where you can't prefix env vars.

getchoo:
I'll admit I definitley went overboard on the default config, but what options would you say should stay there? I was thinking of removing most of the `*flags` options, but keeping some like `PreBuildCommand`, `MakepkgConf`, `Pager`, `Redownload`, and `NoCheck` - since they would be likely used by at least some people (or at least by me with `MakepkgConf` for chroot builds).

I'll also add makepkgconf to the man page in a commit soon; i'm just a bit busy with finals

Morganamilo:
The config should stay as is. There's a man page that lists all the options and people should read that.

getchoo:
...I think I just accidentally pushed a horrible branch. Going to fix now

getchoo:
Sorry about that, should be good now.

man/paru.8
```diff
@@ -250,11 +250,21 @@ Directory used to download and run PKGBUILDs.
...
+.B \-\-makepkgconf <file>
+Specifies a \fBmakepkg.conf\fR file to use in the chroot enviorment. The file
+can only be an absolute path to the file.
+
 .TP
 .B \-\-pacman <command>
 The command to use for \fBpacman\fR calls. This can be a command in
 \fBPATH\fR or an absolute path to the file.
 
+.TP
```
Morganamilo, line 263: This is a pacman flag and hence isnt documented in paru's man page.

getchoo:
fixed

PR #606, opened 2021-12-05
Update spanish translation

alosarjos:


alosarjos:
Personally I think some strings should start with uppercase like on this case the marked out of date:

image

Morganamilo:
I think there was something about that previously. Can't remember what happened to the discussion.

po/es.po
```diff
@@ -186,356 +194,742 @@ msgstr "marcado como desactualizado: "
...
+#: src/help.rs:82
+msgid ""
+"    --topdown              Shows repository's packages first and then AUR's"
+msgstr ""
+"    --topdown              Muestra primero los paquetes de los repositorios "
+"y después de los AUR"
+
+#: src/help.rs:84
+msgid "show specific options:"
+msgstr "mostrar opciones específicas:"
```
Morganamilo, line 599: show is the name of the argument: `--show`

alosarjos: Updated to mean something more alike to that. Thanks for checking it, without context I couldn't know.

Morganamilo: It's changed but --show still seems to be translated.

alosarjos: You want "show" not to be translated? Now it means something like "specific options of view" so it's clear it's refering to the specific options of the looks, but I can leave it as "show" without translation if you prefer

Morganamilo: It makes more sense in the context of the help menu 

```
New operations:
    paru {-P --show}        [options]
    paru {-G --getpkgbuild} [package(s)]

show specific options:
    -c --complete         Used for completions
    -s --stats            Display system package statistics
    -w --news             Print arch news

getpkgbuild specific options:
    -p --print            Print pkgbuild to stdout
    -c --comments         Print AUR comments for pkgbuild

upgrade specific options:
    -i --install          Install package as well as building
```

alosarjos: That should be better now, I also changed the "upgrade specific options" in the same way

PR #607, opened 2021-12-05
update russian translation

alex19EP:
I made the translation more like pacman and the AUR web interface.
also fixed some oddities here and there.

### DIFF
--- a/po/ru.po
+++ b/po/ru.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: paru VERSION\n"
 "Report-Msgid-Bugs-To: https://github.com/Morganamilo/paru\n"
-"POT-Creation-Date: 2021-06-28 03:24+0000\n"
+"POT-Creation-Date: 2021-12-05 07:35+0000\n"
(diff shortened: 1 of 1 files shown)


po/ru.po
```diff
@@ -186,374 +195,710 @@ msgstr "помечены как устаревшие: "
...
 
-#: src/fmt.rs:6 src/info.rs:146 src/search.rs:176
+#: src/fmt.rs:6 src/info.rs:151 src/search.rs:214
 msgid "None"
 msgstr "Никаких"
 
-#: src/info.rs:30 src/query.rs:51
+#: src/help.rs:4
+msgid "Usage:"
+msgstr ""
```
Morganamilo, line 250: None of these have been translated

alex19EP: indeed. I am too lazy to do it now🙂.

can you merge this and I will translate --help with my girlfriend on next weekend.

Morganamilo: Right right if it's intentional that's okay, thought you may have missed it. I'll assume the rest is all good then.

PR #628, opened 2021-12-18
Install "base" package when creating chroot

MithicSpirit:
Fixes builds that assume dependencies of base will be installed (which is reasonable since any real Arch system will have it installed), such as those using nvm.


Morganamilo:
Base is not required, if a package makes the assumption they need to be fixed.

PR #634, opened 2021-12-23
Add note about hotediting?!

boehs:


Morganamilo:
Idea's good though I'd word it more like "The PKGBUILDs may then be edite". Maybe also note the edit's wont presisit unless the changes are commited.

Issue #304, opened 2021-04-02
Allow users to build AUR packages from root with a flag


kotx:
**Have you checked the readme and man page for this feature?** yes
**Have you checked previous issues for this feature?** yes
~~I know about all the security risks and whatever, but sometimes those don't apply. There should be a flag to allow users to build AUR packages as root.~~

~~Sometimes what the user wants is more than what is best practice ;)~~
Edit: Since building as root is a bad idea, it would be a good convenience feature to automatically de-escalate and build as a different user, then install files escalated.

trevorgunn:
Packages are always installed as root...

trevorgunn:
If you mean *building* a package as root then that flag won't do anything because makepkg (which paru calls) will still not let you build as root.

kotx:
> If you mean _building_ a package as root then that flag won't do anything because makepkg (which paru calls) will still not let you build as root.

Yeah, that's what I meant. It's possible to [patch](https://smjrifle.net/restore-asroot-yaourtmakgpkg-arch/) makepkg to allow execution as root, but it doesn't seem to be as easy in the case of paru.

trevorgunn:
Well you would patch paru the same way. But if you're going to go through all this trouble to run makepkg as root, why not go through the trouble to learn how not to run makepkg as root? Surely it's a similar amount of effort.

http://allanmcrae.com/2015/01/replacing-makepkg-asroot/

kotx:
> Well you would patch paru the same way. But if you're going to go through all this trouble to run makepkg as root, why not go through the trouble to learn how not to run makepkg as root? Surely it's a similar amount of effort.
> 
> http://allanmcrae.com/2015/01/replacing-makepkg-asroot/

Well surely it would be faster to type `paru -f ...` with a patched makepkg than git cloning from the AUR and using makepkg with that. I guess this is a niche use case.

Maybe paru could run as its own user if it detects it's being run as root? It doesn't sound ideal though.

koraa:
Why doesn't paru just use a dedeicated `parubuild` user to compile packages?

soloturn:
@kotx why you want to build a package as root user?

koraa:
E.g. because I am writing a dockerfile and wish to avoid the complexity of creating a sudoers ffile…

kotx:
Sometimes I want to install a package while setting up a new system that I know is 100% safe, and it's a hassle switching between users.

koraa:
[...]

PR #665, opened 2022-01-29
Add container feature: Compilation version to run paru as root

vejnar:
See #304


vejnar:
Rebased.

vejnar:
I have been using this feature in containers for a while now without problem. I would appreciate your feedback. Please let me know if you intent to accept this request.

PR #668, opened 2022-02-01
Option to keep old package versions with local repo

JeWe37:
**Motivation:** I want to use the local repository functionality to keep a record of all versions of AUR packages that I have installed, as there is no equivalent of say the ALA. The git versions of the PKGBUILDs have the drawback of depending on potentially dead external resources. I often find myself clearing the regular cache though and it is far less feasible to centralize it(say for faster installs of AUR packages already installed on my PC on my Laptop). 

There might very well be further applications for this that I have not thought of but this is my use case.

**Changes:** Therefore I would like to keep a record of the packages in the local repository, however this was prevented by the `-R` option being set on `repo-add`. This PR adds a simple set of CLI and config options to optionally disable that flag.


Morganamilo:
Looks good but don't know about the name. How about `KeepDbPackages`? `KeepRepoPackages`?

JeWe37:
Can't say I have strong opinions on the naming, `KeepDbPackages` seems slightly off since the DB is the part that isn't, so how about I go with `KeepRepoPackages`? Any preferences as to what I should do with the command line arguments or keep those as-is?

Morganamilo:
KeepRepoPackages slightly conflicts because most user's will probably not realise it's to do with local repos. KeepLocalRepoPackages would be too long too.

Ideally the config and flag names would be the same. Some times they's a bit different when it's useful for the flag to be short (like --aur) but then lacks a bit of context.

Not gonna bike shed forever though KeepRepoPackages is fine.

Actually KeepRepoCache may be better actually. It's up to you which sounds better.

JeWe37:
`KeepRepoCache` it is. Alongside `--keeprepocache` and `--nokeeprepocache` then. I should be able to find the time to finish this off this weekend. I saw there were some changes with a new command interface but that shouldn't be a big deal.

man/paru.8
```diff
@@ -591,6 +591,15 @@ Sign packages with gpg. Optionally indicate which key to sign with.
 .B \-\-nosign
 Don't sign package with gpg.
 
+.TP
+.B \-\-keeprepocache [= key]
```
Morganamilo, line 595: `[= key]` copy paste error?

JeWe37: Yep... fixed

PR #669, opened 2022-02-02
Implement flag to download AUR packages via SSH

### LINKED ISSUE #399: Clone using ssh instead of https
`paru -G mypackage` will clone the package from AUR using `https`.

I'd like for this to be done via `ssh`, so I can also push changes for packages pulled this way.

WhyNotHugo:
Implement a new `-e` or `--editable` flag to download AUR packages using
SSH.

Fixes #399

### DIFF
--- a/src/command_line.rs
+++ b/src/command_line.rs
@@ -203,7 +203,11 @@ impl Config {
             Arg::Long("searchby") => self.search_by = ConfigEnum::from_str(argkey, value?)?,
             Arg::Long("limit") => self.limit = value?.parse()?,
             Arg::Long("news") | Arg::Short('w') => self.news += 1,
-            Arg::Long("stats") | Arg::Short('s') => self.stats = true,
+            Arg::Long("stats") => self.stats = true,
+            Arg::Short('s') => {
+                self.stats = true;
+                self.ssh = true;
+            }
             Arg::Long("order") | Arg::Short('o') => self.order = true,
             Arg::Long("removemake") => {
                 self.remove_make = YesNoAsk::Yes.default_or(argkey, value.ok())?
@@ -269,6 +273,7 @@ impl Config {
             Arg::Long("newsonupgrade") => self.news_on_upgrade = true,
             Arg::Long("nonewsonupgrade") => self.news_on_upgrade = false,
             Arg::Long("comments") => self.comments = true,
+            Arg::Long("ssh") => self.ssh = true,
             // ops
             Arg::Long("database") | Arg::Short('D') => set_op(Op::Database),
             Arg::Long("files") | Arg::Short('F') => set_op(Op::Files),
--- a/src/config.rs
+++ b/src/config.rs
@@ -390,6 +390,7 @@ pub struct Config {
     pub print: bool,
     pub news_on_upgrade: bool,
     pub comments: bool,
+    pub ssh: bool,
     pub sign: Sign,
     pub sign_db: Sign,
 
@@ -614,12 +615,22 @@ impl Config {
             self.raur = crate::mock::Mock::new()?;
         }
 
+        let aur_url = if self.ssh {
+            self.aur_url
+                .to_string()
+                .replacen("https://", "ssh://aur@", 1)
+                .parse()
+                .expect("change AUR URL schema from HTTPS to SSH")
+        } else {
+            self.aur_url.clone()
+        };
+
         self.fetch = aur_fetch::Handle {
             git: self.git_bin.clone().into(),
             git_flags: self.git_flags.clone(),
             clone_dir: self.build_dir.clone(),
             diff_dir: self.cache_dir.join("diff"),
-            aur_url: self.aur_url.clone(),
+            aur_url,
         };
 
         self.need_root = self.need_root();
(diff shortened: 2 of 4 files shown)


WhyNotHugo:
Oh, the workaround mentioned in #399 is problematic when updating the system since all those clones are done with SSH and require authorising the agent.

Morganamilo:
> Oh, the workaround mentioned in https://github.com/Morganamilo/paru/issues/399 is problematic when updating the system since all those clones are done with SSH and require authorising the agent.

Are you saying it will fail to pull over ssh on an aur repo you don't own?

WhyNotHugo:
> Are you saying it will fail to pull over ssh on an aur repo you don't own?

No, that's not it.

The issue with the override in #399 is that _all_ AUR clones are done via SSH, so if I run something like `paru -Syu`, it also tries to clone via SSH and starts prompting for my password (or for the ssh agent to unlock).

I also intend on using a hardware-backed (e.g.: yubikey) SSH keys for the AUR as soon as support for that is deployed, but that won't help much; I'll still need to tap the key for each clone.

In reality, I don't want to clone via SSH _most_ of the time, only in exceptional situations when I want to make changes and push them back.

Morganamilo:
Right that sounds reasonable, though i'd rather the name be --ssh. Also you hard coded the url. you should instead take url that is configured.

WhyNotHugo:
> Right that sounds reasonable, though i'd rather the name be --ssh.

:+1: 

> Also you hard coded the url. you should instead take url that is configured.

Do you mean a new config setting with the base SSH url? `AurSshUrl`?

Morganamilo:
No new section. just take the normal aururl and change the protocol to ssh.

src/config.rs
```diff
@@ -614,12 +615,22 @@ impl Config {
             self.raur = crate::mock::Mock::new()?;
         }
 
+        let aur_url = if self.ssh {
+            self.aur_url
+                .to_string()
```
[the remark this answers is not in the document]
WhyNotHugo:
```
Do nothing and return `Err` under the following circumstances:

[...]

* If either the old or new scheme is `http`, `https`, `ws`,
  `wss` or `ftp` and the other is not one of these
```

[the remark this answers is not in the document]
WhyNotHugo: I've no idea why this restriction is there, but it means we can't use `set_scheme`.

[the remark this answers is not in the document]
Morganamilo: Sad https://github.com/servo/rust-url/issues/577

Guess this is the way for now.

src/command_line.rs
```diff
@@ -269,6 +269,7 @@ impl Config {
             Arg::Long("newsonupgrade") => self.news_on_upgrade = true,
             Arg::Long("nonewsonupgrade") => self.news_on_upgrade = false,
             Arg::Long("comments") => self.comments = true,
+            Arg::Long("ssh") | Arg::Short('e') => self.ssh = true,
```
Morganamilo, line 272: e -> s

WhyNotHugo: (so is `h`)

Morganamilo: Right you need to move 's' to it's own line and set both. see how 'c' is.

Morganamilo: Yeah that's how the code works ATM for the sake of lasiness

WhyNotHugo: Yeah, I realised that pacman's flags is a bit unusual in that it doesn't use typical subcommands.

WhyNotHugo:
I think all comments have been resolved and CI is green. Thanks for all your patience on this one.

src/config.rs
```diff
@@ -614,12 +615,22 @@ impl Config {
...
+                .replacen("https://", "ssh://aur@", 1)
+                .parse()
+                .expect("change AUR URL schema from HTTPS to SSH")
```
Morganamilo, line 623: use ? instead

src/help.rs
```diff
@@ -89,6 +89,7 @@ pub fn help() {
     printtr!("getpkgbuild specific options:");
     printtr!("    -p --print            Print pkgbuild to stdout");
     printtr!("    -c --comments         Print AUR comments for pkgbuild");
+    printtr!("    -e --ssh              Fetch using a read-write remote (e.g.: via ssh)");
```
Morganamilo, line 92: just "Clone over ssh" or something would be sufficient. There's no need to abstract away that ssh is used under the hood. If you want the detail that it can be written to then that can be in the man page.

WhyNotHugo:
Also updated commit message and man page.

WhyNotHugo:
Thanks!

RubenKelevra:
@WhyNotHugo wrote:

> > Are you saying it will fail to pull over ssh on an aur repo you don't own?
> 
> No, that's not it.
> 
> The issue with the override in #399 is that _all_ AUR clones are done via SSH, so if I run something like `paru -Syu`, it also tries to clone via SSH and starts prompting for my password (or for the ssh agent to unlock).
> 
> I also intend on using a hardware-backed (e.g.: yubikey) SSH keys for the AUR as soon as support for that is deployed, but that won't help much; I'll still need to tap the key for each clone.
> 
> In reality, I don't want to clone via SSH _most_ of the time, only in exceptional situations when I want to make changes and push them back.

I mean you can use SSH connection multiplexing for the AUR, this should reuse the authentication :)

PR #711, opened 2022-03-31
Improve default settings for paru.conf

AdisonCavani:
It might be a good idea to add a disabled-by-default option for skipping review process into ``paru.conf``.
Can you consider also changing the default option and enable ``BottomUp`` by default?
Displaying the most relevant results at the bottom will improve initial UX.


Morganamilo:
Skipping review is something I very much don't recommend so it's not in the example config.

As for bottom up. Prints stuff in order is how every other tool tends to work including pacman. And paru aims to be as close to pacman as possible.

AdisonCavani:
Ok. Thank you for clarifying

PR #723, opened 2022-04-12
Add armv7h builds

jclds139:
Adds arm7 builds to the workflows, closing #83 

Since [ALARM no longer supports `arm` or `armv6h`](https://archlinuxarm.org/forum/viewtopic.php?f=3&t=15721), this should be the only arm binary still necessary.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,7 @@ jobs:
           [
             "linux/amd64 x86_64",
             "linux/arm64/v8 aarch64",
+            "linux/arm/v7 armv7h"
           ]
     name: Build ${{ matrix.arch }}
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


Morganamilo:
Works. Thanks for finding the fix.

Morganamilo:
Build seems to be broken' any idea what's up?

Morganamilo:
I actually figured it out to be cargo downloading the index filling memory and getting killed by oom. Not sure what caused this change as it's the same error rerunning the .10 release.

Fixed with a nightly flag for now.

Morganamilo:
It may be due to some change in cargo but I don't really want to bisect it. If i do bisect it, it just means Paru will have to build on an old rust version in CI. A version that paru may no longer support. The nightly flag will do for now.

Morganamilo:
You'd think but it still broke when I reverted that.

jclds139:
It looks like this is a known issue with cargo: rust-lang/cargo#10583, and possibly the fault of libgit2/libgit2. Conveniently, libgit2 is also the culprit for the size mismatch on 32bit platforms. We can bypass it by exporting `CARGO_NET_GIT_FETCH_WITH_CLI=true` before running `cargo` as long as `git` is installed in the build container.

PR #724, opened 2022-04-13
Update Dutch translation

flipflop97:
The Dutch translation has multiple sentences that are poorly worded or just plain wrong. This should fix most of these, as well as update some terms to match the Dutch AUR version.

### DIFF
--- a/po/nl.po
+++ b/po/nl.po
@@ -10,8 +10,8 @@ msgstr ""
 "Project-Id-Version: paru VERSION\n"
 "Report-Msgid-Bugs-To: https://github.com/Morganamilo/paru\n"
 "POT-Creation-Date: 2021-06-28 03:24+0000\n"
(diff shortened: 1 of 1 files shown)


po/nl.po
```diff
@@ -609,11 +609,11 @@ msgstr "Totaal aantal geïnstalleerde pakketten: {}"
 
 #: src/stats.rs:87
 msgid "Aur packages: {}"
-msgstr "Aur pakketten: {}"
+msgstr "AUR-pakketten: {}"
```
Morganamilo, line 612: Is there a language reason for adding the hyphen?

flipflop97: Yes, writing compounds as two words is actually a very common mistake in the Netherlands, but is definitely wrong. In Dutch, compound words are by default written as one word. When pronunciation becomes unclear as one word, or when using loan words or abbreviations, the compound is connected using a hyphen. This is an example of the latter.

PR #736, opened 2022-04-20
fix: typo in fish completion

NextAlone:
fish can't complete `paru --s` to `paru --sync` caused by this

* I didn't test all cases

### DIFF
--- a/completions/fish
+++ b/completions/fish
@@ -5,7 +5,7 @@
 set -l progname paru
 
 # paru constants
-set -l noopt 'not __fish_contains_opt -s -s G -s V -s P -s S -s D -s Q -s R -s U -s T -s F database query sync remove upgrade deptest files version'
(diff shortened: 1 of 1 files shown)


NextAlone:
and if we need these two lines, conflict with lines below
```
complete -c $progname -n "$noopt" -a "-D" -d "Modify the package database"
complete -c $progname -n "$noopt" -a "-Q" -d "Query the package database"
```

https://github.com/Morganamilo/paru/blob/master/completions/fish#L31-L32

Morganamilo:
Looks like this had been here for 2 years. Whoops. Thanks for finding.

Morganamilo:
> and if we need these two lines, conflict with lines below
> 
> ```
> complete -c $progname -n "$noopt" -a "-D" -d "Modify the package database"
> complete -c $progname -n "$noopt" -a "-Q" -d "Query the package database"
> ```
> 
> https://github.com/Morganamilo/paru/blob/master/completions/fish#L31-L32

Not quite what you're saying here though.

PR #738, opened 2022-04-23
Fix typo in import GPG keys string

SanchithHegde:
This PR removes an extra bracket added in commit [5b1ff38b0357844193ca7accd2fcf6b13fb4802f](https://github.com/Morganamilo/paru/commit/5b1ff38b0357844193ca7accd2fcf6b13fb4802f#diff-dd5ba1bca5a2e982ec6f7d24bac5d1b5588e12912d838585931ce9547411cccbR47).

PS: I used `sed` to fix the strings, let me know if I have to use `xtr` instead for the translation files.


Morganamilo:
There's a conflict now, just a line number change. Looks good though.

PR #745, opened 2022-05-08
update github bugreport and issue template

RubenKelevra:
- include script to find the currently used paru config file and print it's content minimized
- include script to print the pacman.conf minimized
- cleanup the template a little
- add tags to issues and bugs


RubenKelevra:
@Morganamilo have you seen this? :)

Morganamilo:
I believe I skimmed it a while ago and it was just the opposite direction of what I want.

I don't want users to have to run any command, it's not very hard to copy/drag in a few files.

I also want the template to be very minimal.

PR #749, opened 2022-05-12
fix: `-L -C` detection and incorrect inherited from yay

NextAlone:
- fix: lost `-L --repoctl` detection in fish comp
- fix: also for `-C --chrootctl
- fix: incorrect completion inherited from yay


Morganamilo:
Not a fish user but seems logical. Assuming it's all good.

PR #756, opened 2022-05-16
fix: chrootctl --install and complete with packages

NextAlone:


Morganamilo:
That command will only work for repo packages so aur packages should not be listed.

NextAlone:
> That command will only work for repo packages so aur packages should not be listed.

ok i have fixed it

PR #787, opened 2022-06-28
fix: missing format parameters

zigmars:
Not sure what I am doing (possibly `*_len` format width specifiers given wrong values or not needed at all), 
but this fixed my compile error.

### DIFF
--- a/src/fmt.rs
+++ b/src/fmt.rs
@@ -322,7 +322,10 @@ pub fn print_install_verbose(config: &Config, actions: &Actions) {
                     .map(|pkg| pkg.version().as_str())
                     .unwrap_or(""),
                 pkg.pkg.version().as_str(),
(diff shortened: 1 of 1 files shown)


Morganamilo:
Did you update rust like I said? What version of rust are you using?

zigmars:
Probably I jumped the gun here.
I do not know which version of rust the build system used (but I see on my system only `rustup 1.24.3`, `rustc 1.57.0`).

The full story is, that I tried doing an upgrade of `paru` via `paru -Syu` (on Manjaro-Sway) and it tried to build itself and failed with some errors, which I suppressed by an edit as in this commit; doing the file edit/patch mid-build and then the build succeeded.

Though, I reinstalled now `paru` and can't reproduce any of the errors now. This might be more of an issue for people upgrading to `1.11.0-1`.

Issue #782, opened 2022-06-26
Avoid sycn'ing packages when building


WhyNotHugo:
I'm running `paru -U --localrepo --chroot` locally very often, and I find that running the equivalent of `pacman -Sy` each time is slow and impractical. There's no need to sync each time that operation is run.

I find that running the above is a super useful replacement for just `makepkg -fsc`, but more deterministic.

Could paru stop doing a sync operation when running this command, and have an optional `-y` flag to actually do a sync?

---

BTW: I tried to post this as a discussion, but it seems impossible:

![image](https://user-images.githubusercontent.com/730811/175831279-1ab7ddbe-26f1-4495-bc0a-81e4f1622a87.png)

Morganamilo:
Maybe #764

Morganamilo:
Actually as mkchrootpkg has a -u flag maybe follow that,

WhyNotHugo:
My db gets sync'd twice: once on my host, and another time on the chroot. #764 proposes dropping the second sync. I'd like to opt-out of the first sync too, since I'm just tweaking a local PKGBUILD to make it work, I've no need to update anything from the network.

WhyNotHugo:
> Actually as mkchrootpkg has a -u flag maybe follow that,

No objection on my side.

WhyNotHugo:
`-y` seems like the right choice: https://github.com/Morganamilo/paru/blob/master/src/command_line.rs#L267-L268:

```rust
            Arg::Long("sysupgrade") | Arg::Short('u') => self.sysupgrade = true,
            Arg::Long("refresh") | Arg::Short('y') => self.refresh = true,
```

Morganamilo:
The chroot does a full pacman -Syu, the packages are probably already up to date. -u is the makpkg flag so it fits to match.

WhyNotHugo:
Oh, so behaviour has changed (improved) on master (I think it's due to c60d17eb88be3cfadffc46a2c2493e1e46f01cba).

v1.10.0:

```
> paru -U
:: Resolving dependencies...
🔐 Confirm user presence
:: Synchronizing package databases...
 testing is up to date
 core is up to date
 extra is up to date
 community-testing is up to date
 community is up to date
 multilib-testing is up to date
 multilib is up to date
 paru is up to date
 quarry is up to date
:: Starting full system upgrade...
 there is nothing to do
🔐 Confirm user presence
[...]

PR #789, opened 2022-06-28
Don't update package DB unless `-u` is specified

WhyNotHugo:
When running `paru -U` repeatedly on the same file, paru will update the
system package DB each time. This is slow and impractical when doing
repeated builds of the same package (e.g.: when using `paru` as an
improved `makepkg -fc`).

This change avoids updating the system package DB unless `-u` has been
specified.

Fixes: https://github.com/Morganamilo/paru/issues/782

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -754,7 +754,12 @@ impl Installer {
             if !self.chroot.exists() {
                 self.chroot.create(config, &["base-devel"])?;
             } else {
-                self.chroot.update()?;
+                if config.refresh {
+                    self.chroot.run(&["pacman", "-Sy", "--noconfirm"])?;
+                }
+                if config.sysupgrade {
+                    self.chroot.run(&["pacman", "-Su", "--noconfirm"])?;
+                }
             }
         }
 


WhyNotHugo:
This _mostly_ works, but running `paru -Uu` results in paru printing:

```
:: packages not in the AUR: XXXXX
```

Where `XXXXX` is a list of all my installed non-repo packages... I'm not entirely sure what that is happening.

WhyNotHugo:
Oh, actually this weird warning also shows up on master. However, since `-u` is always implied, it's not an issue. But I think it needs to be fixed for this change to be acceptable.

Morganamilo:
https://github.com/Morganamilo/paru/blob/master/src/install.rs#L1920

WhyNotHugo:
Rebased onto current master. Now `paru -U` will build without syncing the DB each time. `paru -Uu` update the DB as intended, but it is also rebuilding out-of-date AUR packages and I'm not sure why that is happening.

Morganamilo:
That would probably be because -U is no longer it's own thing and actually well integrated with the normal build process.

Morganamilo:
So basically -S and -U is the same function. So all the -S flags kind of apply in the same way.

WhyNotHugo:
So it would make sense for me to use `-y` here, otherwise `-u` is overloaded.

Morganamilo:
Nah. just set sysupgrade to false before the  call to install in the -U path. And set some other flag to true.

WhyNotHugo:
So when using `paru -U`, refresh = `-u`, but when using `paru -S`, refresh = `-y`?

Why not just use the `refresh` flag for both cases?

Morganamilo:
Because -y is for refreshing databases. The chroot update effectively does -Syu inside the chroot. So it does a -y and a -u.

WhyNotHugo:
What do you think of the functionality of the latest version of this PR?

I know the code could be refactored to be cleaner, but I mostly want thoughts on the idea itself.

Morganamilo:
I'd rather one flag to avoid partial upgrades and because `makechrootpkg` only has a -u flag

Morganamilo:
I've come to the realisation that the repo does need to be updated on each run to get access to the new local repo packages.

It would be possible to write some code to only update the local dbs but that would take a bit of effort as we're no longer using a built in flag.

Morganamilo:
So to recap, a pacman -Sy is needed on every chroot build so that the aur packages that were added to the local repo can be picked up. To avoid partial upgrades -u has to come alone too.

Though now, we copy the databses into the chroot so they don't have to be fetched if up to date. And the main chroot updates once before any builds, then each build updates but should have no actual updates due to just being updated. Is this sufficient for you? Seems like the best we're going to get.

PR #816, opened 2022-07-30
Added a way to see the cache size

Devarth123:
# Demo 

it will now also display the cache size when `paru -Ps` is ran


Devarth123:
image

works without any panics now :smiley:

PR #829, opened 2022-08-13
Update sv.po

bittin:
Update Swedish Translation 96% done


Morganamilo:
Looks good, any issue with the final 4%?

bittin:
Nope, just me that did not find the right words too put for them,
maybe some other Swedish person could help

PR #838, opened 2022-08-29
German translation update

SebinNyshkim:
This pull request aims to add missing translations (e.g.: `paru -h`) and update existing ones with tweaked wordings and punctuation.


Morganamilo:
Sorry It's been a while. There's conflicts now as the first commit touches all the translation files (maybe the instructions should be changed to only do the one file. Anyway looks good from what I can see, just need to fix the one conflict.

PR #839, opened 2022-09-04
Don't skip review if --review was explicitly set, even if --noconfirm is set

dscottboggs:
I thought this would make it so `paru -Syu --noconfirm --review` would skip confirmation at pacman updates but still stop for review of AUR packages.

This change doesn't do that though. It still skips over the review process.

Any suggestions on how to get it to work as expected?

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -1572,7 +1572,7 @@ fn print_dir(
 }
 
 fn review(config: &Config, fetch: &aur_fetch::Fetch, pkgs: &[&str]) -> Result<()> {
-    if !config.no_confirm {
+    if !config.no_confirm || !config.skip_review {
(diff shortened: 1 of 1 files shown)


Morganamilo:
Don't see how this is beneficial. Paru already aims to minimise interaction. The one real prompt is the confirmation after review.

dscottboggs:
- `paru` or `paru -Syu` or `paru -Syu --review` leave you to press `Y` before installing main-repo packages.
- `paru -Syu --noconfirm --review` skips both main-repo package confirmation and review

It seems to me that this means that `--review` doesn't do what it's documented to -- unconditionally trigger a review. It does so, but only if _not_ used in conjunction with `--noconfirm`.

PR #841, opened 2022-09-07
Use `terminal_size` crate instead of unmaintained `term_size`

ohno418:
`term_size` crate is now unmaintained. This PR replaces it with `terminal_size` as the maintainers say.

https://github.com/clap-rs/term_size-rs

> This crate is no longer maintained. It works fine, but the maintainers don't have time and willingness to work on it or review future PRs anymore.
> 
> Consider using [terminal_size](https://github.com/eminence/terminal-size/) instead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -269,7 +269,7 @@ dependencies = [
  "encode_unicode",
  "libc",
  "once_cell",
- "terminal_size",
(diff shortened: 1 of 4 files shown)


ohno418:
This seems to be fixed by https://github.com/Morganamilo/paru/pull/1049.

PR #844, opened 2022-09-08
add option/setting "AurRpcUrl"

moson-mo:
Allows to set an alternative URL for the /rpc endpoint.

### DIFF
--- a/completions/fish
+++ b/completions/fish
@@ -189,6 +189,7 @@ complete -c $progname -n "$chrootctl" -s u -l sysupgrade -d 'Upgrade the chroot'
 complete -c $progname -n "not $noopt" -l repo -d 'Assume targets are from the repositories' -f
 complete -c $progname -n "not $noopt" -s a -l aur -d 'Assume targets are from the AUR' -f
 complete -c $progname -n "not $noopt" -l aururl -d 'Set an alternative AUR URL' -f
+complete -c $progname -n "not $noopt" -l aurrpcurl -d 'Set an alternative URL for the AUR /rpc endpoint' -f
 complete -c $progname -n "not $noopt" -l clonedir -d 'Directory used to download and run PKGBUILDs' -r
 complete -c $progname -n "not $noopt" -l makepkg -d 'Makepkg command to use' -f
 complete -c $progname -n "not $noopt" -l pacman -d 'Pacman command to use' -f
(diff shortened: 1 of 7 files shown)


Morganamilo:
The option needs to fall back to the aur when not set. With how it is currently, a user wil set --aururl and not realize they are still using the aur's RPC end point.

This can be achieved by. Making the option an `Option<URL>` and doing `self.aur_rpc_url.or_else(|| self.aur_url.join("RPC")

Morganamilo:
Whoops fat fingered the close button. Mobile :(

moson-mo:
Thanks for you fast reply

> The option needs to fall back to the aur when not set.

Good point.
I've tried to address that in another commit.

I'm not really familiar with rust I must say, so I hope that makes some sense what I did there 😆 
At least I could make the compiler happy 😉 

But my tests on that were looking fine:

```
$ target/debug/paru -Ss paru --aururl https://google.com                                                                                                                                                                                                        
error: aur search failed: HTTP status client error (404 Not Found) for url (https://google.com/rpc)
$ target/debug/paru -Ss paru --aururl https://google.com --aurrpcurl https://server.moson.rocks/rpc                                                                                                                                                           
aur/paru 1.11.1-1 [+596 ~29.82]
    Feature packed AUR helper
aur/paru-bin 1.11.1-1 [+184 ~12.33]
    Feature packed AUR helper
aur/paru-git 1.7.3.r22.g1ed1f29-1 [+21 ~2.53]
    Feature packed AUR helper
aur/paruz 1.1.2-1 [+12 ~0.80]
    A fzf terminal UI for paru or pacman
aur/lsparu 1.0.0-1 [+6 ~0.37]
    A fzf TUI for paru
aur/parui-git 0.1.14.r0.e79f9bb-3 [+3 ~0.31]
    Simple TUI frontend for paru.

```

Morganamilo:
Looks okay, though I guess with the possible error from join making it a bit awkward it's probably best to just do it as a match 

```rust
            let rpc_url = match &self.aur_rpc_url {
                Some(rpc) => rpc.to_string(),
                None => self.aur_url.join("rpc")?.to_string(),
            };

            self.raur = raur::Handle::new_with_settings(client, rpc_url);
```

Other than that it should be good.

moson-mo:
Ok cool. Changed now.

PR #853, opened 2022-09-25
Don't try to parse devel.json if it's empty

guihkx:
In rare scenarios with low disk space left, running `paru -Syu --devel` could cause an abrupt interruption, due to 'No space left on device (os error 28)':

image

In that scenario, the original `devel.json` could be replaced by an empty file, which would cause `paru` to fail on future `--devel` updates, due to not being able to parse an empty file as json:

image

This fixes the issue by simply ignoring `devel.json` when it's empty.


guihkx:
Actually, this isn't the proper fix and I think I found the fix for the underlying issue.

PR #854, opened 2022-09-25
Write to a temporary file before updating `devel.json`

guihkx:
Running `paru -Syu --devel` with no disk space left was causing `devel.json` to be replaced by an empty file.

Now, we try to write to a temporary file first (`devel.json.tmp`), and if that fails, the original `devel.json` is kept.


Morganamilo:
Code looks good.

The error string needs to be in tr! to be translatable though. And should be worded like "failed to rename" failed to write"

Also I'm wondering if it would be better to just ignore the file if it's empty.

guihkx:
>The error string needs to be in tr! to be translatable though

I didn't wrap them in `tr!()` because other strings in the same function aren't translatable. Should I make them translatable too?

>Also I'm wondering if it would be better to just ignore the file if it's empty.

That was actually my first approach in #853, but I realized that if we ignore an empty `devel.json` forever, users will never be able to update VCS packages unless they manually remove the empty `devel.json`...

Morganamilo:
> I didn't wrap them in tr!() because other strings in the same function aren't translatable. Should I make them translatable too?

Yeah if any are missing they need to be translated too.

> That was actually my first approach in https://github.com/Morganamilo/paru/pull/853, but I realized that if we ignore an empty devel.json forever, users will never be able to update VCS packages unless they manually remove the empty devel.json...

How so?

guihkx:
Um... Yeah, actually what I said makes no sense. 😅 

However, I'm still highly in favor of preventing the 'breakage' of `devel.json`.

Nevertheless, let me know what you think it's best and, if necessary, I'll reopen my previous PR in favor of this one.

Thanks!

Morganamilo:
Thinking about it, temp file is the better way.

guihkx:
Done! Let me know if the error messages are good enough.

I tried to update the potfiles in the previous rebase I did, but for some reason `de.pot` looked off to me (wrong translation or whatever), and I didn't want to break anything, so I dropped the commit.

src/devel.rs
```diff
@@ -199,16 +199,35 @@ pub async fn gendb(config: &mut Config) -> Result<()> {
...
+        file.with_context(|| tr!("failed to create temporary file: {}", temp.display()))?;
     let json = serde_json::to_string_pretty(&devel_info).unwrap();
-    file.write_all(json.as_bytes())?;
+
+    file.write_all(json.as_bytes())
+        .with_context(|| tr!("failed to write to temporary file: {}", temp.display()))?;
+
+    std::fs::rename(&temp, &config.devel_path).with_context(|| {
+        tr!(
+            "failed to rename '{}' to '{}",
```
Morganamilo, line 226: The args need to be named. e.g.

```rust
        tr!(
                "Generating .SRCINFO for {repo}/{dir}...",
                repo = repo.name,
                dir = path.file_name().unwrap().to_string_lossy()
        );

```

guihkx:
I'm not sure if I'm doing something wrong, but after I ran this:

```
./scripts/mkpot && ./scripts/updpo
```

The script generated some incorrectly translated strings (according to Google Translate, anyway) in `de.po`:

```
#: src/devel.rs:204
#, fuzzy
msgid "failed to create cache directory: {dir}"
msgstr "Cache-Verzeichnis konnte nicht gefunden werden"

#: src/devel.rs:219
#, fuzzy
msgid "failed to create temporary file: {temp}"
msgstr "Dateimanager konnte nicht ausgeführt werden: {}"

#: src/devel.rs:227
#, fuzzy
msgid "failed to write to temporary file: {temp}"
msgstr "Verzeichnis konnte nicht gelesen werden: {}"
```

Morganamilo:
Just don't worry about the translations.

Morganamilo:
Also it was right before with the single arg tr!(). The named args are needed when there's multiple. When there's only one don't name it.

src/devel.rs
```diff
@@ -199,16 +199,35 @@ pub async fn gendb(config: &mut Config) -> Result<()> {
...
-    let mut file = file.with_context(|| format!("open: {}", config.devel_path.display()))?;
+        .open(&temp);
+    let mut file =
+        file.with_context(|| tr!("failed to create temporary file: {}", temp.display()))?;
     let json = serde_json::to_string_pretty(&devel_info).unwrap();
-    file.write_all(json.as_bytes())?;
+
+    file.write_all(json.as_bytes())
+        .with_context(|| tr!("failed to write to temporary file: {}", temp.display()))?;
+
```
Morganamilo, line 223: I should have spotted it the first time but you probably need to close the file before rename so make sure it's flushed and the bytes really were wriitten.

guihkx: Good catch!

Issue #859, opened 2022-10-03
Building in clean chroot fails using devtools 20221002-1


YHNdnzj:
### Affected Version

`paru v1.11.1 +git - libalpm v13.0.1`

### Description

After updating to the latest devtools (`20221002-1`, currently in testing), building in chroot fails during `pkgver()` for `i3status-rust-git`. It seems that the source dir doesn't exist. Downgrading `devtools` to `20220621-1` fixes the issue for me. I'm not sure if paru is to blame, though.

### Output

[paru.conf](https://github.com/YHNdnzj/dotfiles/blob/master/paru/.config/paru/paru.conf)

```sh
# i3status-rust-git-0.30.0.r2917.g51fd9e5d-1-x86_64-pkgver.log
/startdir/PKGBUILD: line 28: cd: i3status-rust: No such file or directory
```

hexchain:
The issue can be reproduced with a simple stub PKGBUILD:

```sh
pkgname=testpkg
pkgver=1.0
pkgrel=1
pkgdesc="Test package"

prepare() {
    cd "$srcdir"
    touch a
    ls -lR
}

build() {
    cd "$srcdir"
    ls -l a
}

package() {
    cd "$srcdir"
    ls -l a
}
```

Running `paru -U` on the PKGBUILD will fail at `build` step due to the missing file `a`.

YHNdnzj:
After some bisecting, https://github.com/archlinux/devtools/commit/1e23bbc164dedf58a62177fb5da861b6895ce63c seems to be the first bad commit.

Morganamilo:
I don't see how that commit could break it :/

alex19EP:
I see that paru execs makepkg like this:
`makechrootpkg -cr /var/lib/aurbuild/x86_64 -D /var/lib/aur -d /var/cache/pacman/pkg/ -d /var/lib/aur/ -- -feA --noconfirm --holdver`

for me problematic option is -e with out -e in -feA all works fine.

YHNdnzj:
Possible fix at [devtools#117](https://gitlab.archlinux.org/archlinux/devtools/-/merge_requests/117)

SamLukeYes:
[devtools#117](https://gitlab.archlinux.org/archlinux/devtools/-/merge_requests/117) was merged and they've tagged a [new release](https://gitlab.archlinux.org/archlinux/devtools/-/releases/20221012).

YHNdnzj:
[...]

PR #867, opened 2022-10-11
install.rs: add minimal workaround for devtools 20221002-1 makechrootpkg bug

justinkb:
this is the backport in case you want to do a quick #859 bugfix release for 1.11.1


Morganamilo:
I know I'm slow on these. Upstream were kind enough to release a fix a day or two later so this is not needed. Thanks though.

PR #350, opened 2021-05-01
Dependabot

jimman2003:
Add Dependabot

PR #885, opened 2022-11-08
Added dependabot

pentamassiv:
Dependabot will check the dependencies and the used Github Actions for updates and will help to keep them up-to-date

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


pentamassiv:
Nevermind, I just noticed you dont like dependabot https://github.com/Morganamilo/paru/pull/350

pentamassiv:
Would you consider dependabot for the Github Actions? Those don't get too many version bumps and are easily forgotten. docker/setup-qemu-action for example is outdated

Morganamilo:
Especially on libraries. Where the actual version used is whatever the binary uses (within semver of course)

Morganamilo:
Oh wait I thought this PR was also on aur-depends. Scratch that point. But yeah deps are generally updated as needed or before release.

Issue #889, opened 2022-11-13
devel.json no such file


telometto:
### Affected Version
**paru -V**
paru v1.11.2.r84.g492f5b1 - libalpm v13.0.2

### Description
**Have you checked previous issues?**
I keep getting
```
error: open: /home/MYUSER/.local/state/paru/devel.json: No such file or directory (os error 2)
```

### Output
**Include the FULL output of any relevant commands/configs**

**Don't cut parts of the input always include the FULL thing**

**paru.conf and pacman.conf are usually always relevant**

```sh
#
# /etc/pacman.conf
#
# See the pacman.conf(5) manpage for option and repository directives

#
# GENERAL OPTIONS
#
[options]
# The following paths are commented out with their default values listed.
# If you wish to use different paths, uncomment and update the paths.
#RootDir     = /
#DBPath      = /var/lib/pacman/
#CacheDir    = /var/cache/pacman/pkg/
#LogFile     = /var/log/pacman.log
#GPGDir      = /etc/pacman.d/gnupg/
#HookDir     = /etc/pacman.d/hooks/
HoldPkg     = pacman glibc
#XferCommand = /usr/bin/curl -L -C - -f -o %o %u
#XferCommand = /usr/bin/wget --passive-ftp -c -O %o %u
#CleanMethod = KeepInstalled
Architecture = auto

# Pacman won't upgrade packages listed in IgnorePkg and members of IgnoreGroup
#IgnorePkg   =
#IgnoreGroup =

#NoUpgrade   =
#NoExtract   =

# Misc options
#UseSyslog
Color
#NoProgressBar
CheckSpace
#VerbosePkgLists
ParallelDownloads = 5
ILoveCandy

# By default, pacman accepts packages signed by keys that its local keyring
# trusts (see pacman-key and its man page), as well as unsigned packages.
SigLevel    = Required DatabaseOptional
LocalFileSigLevel = Optional
#RemoteFileSigLevel = Required

# NOTE: You must run `pacman-key --init` before first using pacman; the local
# keyring can then be populated with the keys of all official Arch Linux
# packagers with `pacman-key --populate archlinux`.

#
# REPOSITORIES
#   - can be defined here or included from another file
#   - pacman will search repositories in the order defined here
#   - local/custom mirrors can be added here or in separate files
#   - repositories listed first will take precedence when packages
#     have identical names, regardless of version number
#   - URLs will have $repo replaced by the name of the current repo
[...]

PR #892, opened 2022-11-17
Create and initialize the devel file

joezak11:
Create and initialize the devel file if it does not exist and the old devel file does not exist as well.
solves #889


Morganamilo:
I don't think this is the issue. Looks like the issue is the create_dir_all in devel.rs was never changed to create state instead of dir. On that note, the create dir in this bit of code should have been create_dir_all too.

joezak11:
Did I understand correctly?

Morganamilo:
Yep looks good to me now thanks.

PR #897, opened 2022-11-17
Update Dutch translation

flipflop97:
Small translation update


Morganamilo:
Hey there's now conflicts because all the line numbers change. In future this can be avoided by committing the update separately before making changes.

PR #899, opened 2022-11-23
Add missing apostrophe in renaming error message

flipflop97:


Morganamilo:
Good find. CI is red because you must have forked between me pushing an formatted commit and fixing it. But it's not really an issue

I'll probably wait until I'm ready to release and see if I can make this fix to all translations at the same time so translators don't need to update.

flipflop97:
Any updates on this?

PR #900, opened 2022-11-26
Change return code for the case when no news is printed

dylanmtaylor:
An exit code of 1 indicates that the command ran with a failure. It successfully determined there was no news, so should use an exit code of 0.

This exit code 1 behavior seems to cause issues with topgrade, see https://github.com/topgrade-rs/topgrade/issues/232 and https://github.com/topgrade-rs/topgrade/issues/220.


Morganamilo:
returning 1 with no results is convention. Pacman does it for -Qu and -Ss for example.

PR #916, opened 2022-12-16
Fix no sudo when running paru -Sc

MithicSpirit:
Fixes an issue where `paru -Sc` would run pacman without root permissions, resulting in
> `error: you cannot perform this operation unless you are root.`

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -773,6 +773,10 @@ impl Config {
             aur_url,
         };
 
+        if self.mode == Mode::empty() {
+            self.mode = Mode::all();
+        }
+
         self.need_root = self.need_root();
 
         if self.repos != LocalRepos::None {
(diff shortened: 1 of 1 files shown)


MithicSpirit:
Not sure why the tests failed... this should only affect it when the `-c`/`--clean` argument is present, and none of the failing tests have that.

Morganamilo:
Have you tested this your self? I believe it should be `!self.mode.repo()`

MithicSpirit:
> Have you tested this your self? I believe it should be `!self.mode.repo()`

Yeah, I thought that at first too, but for some reason it doesn't work if I do that.

MithicSpirit:
Looks like the issue was that the mode hadn't "defaulted" to all yet. Moving that earlier should hopefully fix it properly now.

MithicSpirit:
Considering that the same checks also failed on #917 (which only changes documentation), I suspect that the issue is with the tests themselves and not with this PR.

Morganamilo:
The issue is pacman-git outright broke repo-add lol

Morganamilo:
> Looks like the issue was that the mode hadn't "defaulted" to all yet. Moving that earlier should hopefully fix it properly now.

Sounds right to me. Good find.

PR #950, opened 2023-02-25
Let child pacman process inherit stdio

shurizzle:
`paru` fails when trying to spawn a `sudo pacman` child process (`isatty()` returns false) so I let the child inherit stdin to solve the bug.


Morganamilo:
Can you give output of this bug occuring?

shurizzle:
https://asciinema.org/a/CWfmc6BhotCJHY4CBLnK3kpxt

You can see it on the last line

shurizzle:
To replicate the problem with sudo you need to add `Defaults timestamp_timeout = 0` in your configuration file (ask for password every time) or use `doas` or `pleaser`. Hope this helps to understand better the problem.

PR #993, opened 2023-05-17
Show diffs then ask to edit files, when review is enabled

zer0-x:
The previews behavior was:
```mermaid
flowchart TB
    A([Review is enabled?]) -->|Yes| B(FileManager is specified?)
    B --> |Yes| C(Show FileManager)

    B -->|No| D(Show diff)
    C --> E([Make package])
    D --> E

```
Now it's like that:
```mermaid
flowchart TB
    A([Review is enabled?]) -->|Yes| B(Show diff)
    B --> C(FileManager is specified?)

    C -->|Yes| D(Show FileManager)
    C -->|No| E([Make package])
    D --> E

```

## I did
- Remove an `if statement` and reorder some code.
- Remove one of the `Accept changes?` prompts, so it will not be prompted two times with the new flow.
- Add `Proceed to edit files?` prompt to ask before opening the file manager after viewing the diff.

>  `cargo fmt`  made it harder to see I did.

## Why
It's better to view the diff, then if you want to edit something you can.

It may not be necessary, but might save some time.

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -1681,73 +1681,72 @@ fn review(config: &Config, fetch: &aur_fetch::Fetch, pkgs: &[&str]) -> Result<()
     if pkgs.is_empty() {
         return Ok(());
     }
+
     if !config.no_confirm {
-        if let Some(ref fm) = config.fm {
-            let _view = file_manager(config, fetch, fm, pkgs)?;
+        let unseen = fetch.unseen(pkgs)?;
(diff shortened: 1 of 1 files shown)


Morganamilo:
The directory passed to the file manager contains all the diffs. Why do you also want them output here?

zer0-x:
After thinking for a secund time, now I can see that it is completely unnecessary.

PR #1011, opened 2023-06-29
Update zh_CN translation

clsty:


clsty:
Update `zh_CN.po` translation.
- Update previous translators according to history commits.
- Fix some wrong translations.
- Fix many unsuitable (e.g. inconsistency) translations.
- Translated some untranslated.

PR #1014, opened 2023-07-09
Suggest `SigLevel = Never` when prompting user to add local repo to pacman.conf

chenxiaolong:
Currently, when using the local repo option, paru's help text suggests setting `SigLevel` to `PackageOptional DatabaseOptional`, along with adding the local repo as an additional `CacheDir` to avoid unnecessary copies to `/var/cache/pacman/pkg`. Unfortunately, since pacman 6, this doesn't actually work when packages are unsigned. With optional signatures, if pacman can't find the `.sig` in the cache, it triggers the download logic, causing the local packages to be copied to the first listed `CacheDir`.

Since most users are unlikely to sign their locally-built AUR packages, it probably makes the most sense to have the help text suggest setting `SigLevel` to `Never` instead, which avoids this problem since pacman won't try to look for the `.sig` in the cache. Folks who sign their packages are unaffected regardless of the `SigLevel` value since there's never a situation where a file would be "missing" from the cache.

[1] https://bugs.archlinux.org/task/71109

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -800,7 +800,7 @@ impl Config {
     CacheDir = /var/lib/repo/aur
 
     [aur]
-    SigLevel = PackageOptional DatabaseOptional
+    SigLevel = Never
     Server = file:///var/lib/repo/aur"
                 );
             }


Morganamilo:
> causing the local packages to be copied to the first listed CacheDir.

That may count as a regression in pacman so maybe it should be fixed there. I don't really like the idea of explicitly ignoring sigs.

chenxiaolong:
Fair enough and I agree it should preferably be fixed in pacman. It's a [known issue upstream](https://bugs.archlinux.org/task/71109), though it hasn't gotten much traction.

Morganamilo:
I'll get around to it eventually. Hoping to pick back up on projects the best that I can.

PR #1018, opened 2023-07-23
Update Cargo.toml http => https

szabgab:


szabgab:
Thanks for accepting. I am building the [rust digger](https://rust-digger.code-maven.com/) to find similar simple and also slightly more complex issues with crates.

PR #1024, opened 2023-08-17
Don't forget about git

georgewoodall82:
Doesn't work without git.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Paru is your standard pacman wrapping AUR helper with lots of features and minim
 ## Installation
 
 ```
-sudo pacman -S --needed base-devel
+sudo pacman -S --needed base-devel git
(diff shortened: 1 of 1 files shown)


SandaruKasa:
Since I'm here lurking in the PRs, @Morganamilo, do you mind merging this one? It's really minuscule.

Thanks in advance! No pressure tho

Morganamilo:
The install instructions aren't really meant to be a "copy paste and go" job. Sudo also isn't installed by default but is mentioned there.

The only reason base-devel is mentioned here is because it results in a ton of clueless issue reports otherwise.

Now that base-devel is a meta package it will probably end up a runtime dep of paru and it may avoid all this as long as makeup installers base-devel before it needs it

But in general i expect users to understand how to install packages without help as it's a much needed skill when something doesnt go quite right.

PR #1043, opened 2023-10-11
Update README.md

lollilol:
installation section:
added install "git". Its not preinstalled sometimes.


iruoy:
This should be closed. Somehow a typo snuck into this 4 character commit. Other than that it's the same as #1024

PR #1046, opened 2023-10-12
kuchiki -> scraper

### LINKED ISSUE #931: Kuchiki is unmaintained
This is to let you know that your dependency https://crates.io/crates/kuchiki will not receive new versions. I’ve archived its source repository: https://github.com/kuchiki-rs/kuchiki#archived

No immediate action is required as I’m not planning the yank the crate. If and when you look into replacing the dependency, I’d like to encourage you to experiment with data structures since `Rc`+`Weak` is only one of many ways to represent trees in Rust.

SandaruKasa:
Resolves #931

An API from AUR would, of course, be more convenient, but it is what it is.

Bonus: just like `kuchiki`, `scraper` is just a front-end for `html5ever` & `cssparser` (but a newer version of it), so not much has changed in terms of dependencies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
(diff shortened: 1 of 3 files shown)


SandaruKasa:
Oops

PR #1044, opened 2023-10-11
Fix chrono depricated warnings

balroggg:
Fix `chrono` depricated warnings and clippy warnings.

PR #1047, opened 2023-10-12
Fix deprecated `from_utc`

SandaruKasa:
A no-brainer, really. Even the bodies of those functions are the same. It's just that `from_utc` wasn't verbose enough, so now it's deprecated and they ask you to use `from_naive_utc_and_offset` instead.

Makes cargo shut up about the use of a deprecated function.


SandaruKasa:
Nevermind, somebody already got there faster than me and even fixed most of the clippy warning.

Closing in favor of #1044

PR #1050, opened 2023-10-13
Update README.md

thePanz:
Document how to override the default configurations by using `~/.config/paru/paru.conf`

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,20 +66,7 @@ See [CONTRIBUTING.md](./CONTRIBUTING.md).
 ## Default Configuration
 
 Paru will read `/etc/paru.conf` or `~/.config/paru/paru.conf` (if available) to load the default configuration.
-See the default configuration [paru.conf](./paru.conf) as a reference.
-
-To override the default configuration, create the `~/config/paru/paru.conf` file and configure it according to your needs.
-Example:
-```
-# file: ~/.config/paru/paru.conf
-[options]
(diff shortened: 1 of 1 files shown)


Morganamilo:
Theres already a man page for the config file. I expect users to know how a config files generally work as well.

Plus I don't expect most users to come to this page to install paru as they'd get it through package management.

thePanz:
I agree, the man pages already defines how the paru.conf file is expected to be found and used

The missing point for me was: did not know that paru was using a config file!

Would it make sense to add a note about it (for the distracted users like me)? wdyt?

Morganamilo:
The mention of the config fine is mentioned in the main man page. If people don't check that I doubt an online readme will help.

thePanz:
Fair enough, even if I strongly suggest to have a mention online: this would be the first place to know the features offered by the tool (before installing).

I noticed that other tools (https://github.com/BurntSushi/ripgrep/blob/master/GUIDE.md#configuration-file) mentions that on the online documentation.
While `rg` does not provide a manpage for their `ripgreprc` configuration tho

Morganamilo:
If you just want the config file locations mentioned I think that would make sense.

thePanz:
sure! I will update this PR later on

soloturn:
@thePanz you want to cancel the pr,?

thePanz:
@soloturn thanks for the headsup! I updated the PR now, with the latest comments from @Morganamilo

PR #1071, opened 2023-11-13
Add news_url in config.rs

weitcis:
This PR allows the user to specify which rss feed to fetch news from by specifying a URL after NewsOnUpgrade. This can be useful for users of other pacman-based distros.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -416,6 +416,8 @@ pub struct Config {
     pub aur_rpc_url: Option<Url>,
     #[default(Url::parse("https://archlinux.org").unwrap())]
     pub arch_url: Url,
+    #[default(Url::parse("https://archlinux.org/feeds/news/").unwrap())]
+    pub news_url: Url,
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -1056,7 +1057,12 @@ impl Config {
             "BatchInstall" => self.batch_install = true,
             "UseAsk" => self.use_ask = true,
             "SaveChanges" => self.save_changes = true,
-            "NewsOnUpgrade" => self.news_on_upgrade = true,
```
Morganamilo, line 1059: Seems weird to have 2 ways to set the news url. I think just the direct option is fine.

uetcis: Sorry I didn't get it. What do you mean by the direct option?

uetcis: Do you mean that we should remove `news_on_upgrade` and only leave `news_url`?

Morganamilo: yeah

PR #1074, opened 2023-11-19
Add printing of which package failed to clean in aur cache

triarius:
I recently installed a package from the AUR that `srcinfo.rs` failed to parse the `.SRCINFO` of. As a result, when I attempted to clean the caches (with `paru -Sc`), I get an error message of the parse failure, but there is no indication of which package the error originated from. This PR adds that information to the output.

### DIFF
--- a/src/clean.rs
+++ b/src/clean.rs
@@ -94,16 +94,18 @@ fn clean_aur(
     let cached_pkgs = read_dir(&config.fetch.clone_dir)
         .with_context(|| tr!("can't open clone dir: {}", config.fetch.clone_dir.display()))?;
 
-    #[allow(unused_must_use)]
-    cached_pkgs.for_each(|maybe_pkg| {
-        maybe_pkg.map_err(anyhow::Error::from).map(|path| {
-            clean_aur_pkg(config, &path, remove_all, keep_installed, keep_current, rm).map_err(
-                |err| {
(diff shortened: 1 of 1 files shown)


Morganamilo:
Good change but I'm not a fan of the nesting here and change of a for loop to a for_each. The compiler warning about unused values is a sign it's probably not the ideal way to write this.

I think the way to do this is just .with_context() the original error returned by `clean_aur_pkg` then `if let` it as usual. Also the error message needs to be in a `tr!()` so it can be translated.

triarius:
@Morganamilo how about now?

This is what it looks like

src/clean.rs
```diff
@@ -136,7 +135,13 @@ fn clean_aur_pkg(
         return do_remove(config, &file.path(), rm);
     }
 
-    let srcinfo = Srcinfo::parse_file(file.path().join(".SRCINFO"))?;
+    let srcinfo = Srcinfo::parse_file(file.path().join(".SRCINFO")).with_context(|| {
+        let file_name = file.file_name();
+        tr!(
+            "could not parse .SRCINFO for '{}'",
+            file_name.to_string_lossy()
```
Morganamilo, line 142: use .display() here instead. Otherwise all good.

Morganamilo: Actually this is the filename and not a Path so nevermind.

PR #1141, opened 2024-02-24
Fix cache_dir instead of state_dir in error message

paullouisageneau:
This PR makes the `failed to create state directory` error display the proper directory, as it can be quite confusing when it happens.


soloturn:
mind merging @Morganamilo ?

PR #1160, opened 2024-03-19
scripts/mkmo: switch msgfmt args

ralphptorres:
When POSIXLY_CORRECT is set, msgfmt requires options to be specified before the po files (as also defined in its man).


ralphptorres:
Fixes build errors on archlinux when said env var is set.

PR #1191, opened 2024-05-04
Update README.md

Heus-Sueh:
Adds a command line with chained commands to make life easier for users

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,12 @@ cd paru
 makepkg -si
 ```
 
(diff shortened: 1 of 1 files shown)


CodingWithAnxiety:
I'd consider chaining commands a really bad habit, especially when copy-pasting. You should do things step by step to insure it functionally works on your setup imo. Maybe I'm just odd.

SteavenGamerYT (approved):

PR #145, opened 2020-12-29
Update Installation Instructions

lachlan-smith:
Add git and rustup to installation instructions, as current installation steps do not result in a successful installation on a fresh install using archlinx-2020.12.01-x86_64.iso

PR #1226, opened 2024-08-01
README: Also install git

tyilo:


SandaruKasa:
This has already been suggested at least three times before: #145, #1024, #1043, so maybe it's worth to add at least git to prevent similar reports in future?

tyilo:
If this is not added I think the whole `sudo pacman -S --needed base-devel` line should be removed.

SteavenGamerYT (approved):

ghost (approved):

rien333:
I guess the reasoning behind installing  `base-devel` is that its packages are basically implicitly assumed to be installed in any PKGBUILD on the aur. After all, if you are going to use the aur, you need to be able to build software, and this is precisely what `base-devel` aims to provide (i've seen "bug" reports on the aur that were totally fixeable if the user just had `base-devel`). 

The same cannot be said about git. In fact, `git` it often of an explicit `makedep` of packages that need it.

tyilo:
> I guess the reasoning behind installing `base-devel` is that its packages are basically implicitly assumed to be installed in any PKGBUILD on the aur. After all, if you are going to use the aur, you need to be able to build software, and this is precisely what `base-devel` aims to provide (i've seen "bug" reports on the aur that were totally fixeable if the user just had `base-devel`).
> 
> The same cannot be said about git. In fact, `git` it often of an explicit `makedep` of packages that need it.

Sure. However the next line is literally `git clone https://aur.archlinux.org/paru.git`, which can't be done without having git installed.

PR #1230, opened 2024-08-07
split packages support for GETPKGBUILD

arenevier:
Split packages are not supported in. For example, running `paru -Gp gvim` will display the html content of [this 404 page](https://gitlab.archlinux.org/archlinux/packaging/packages/gvim/-/raw/HEAD/PKGBUILD)

This patch adds support for split packages. In `split_repo_aur_pkgbuilds`,  the base package of the target is checked. If different from the target, the base name will be used instead.

Test: run ```paru -Gp gvim```. It will now display vim PKGBUILD.

No need to do anything for aur. split packages are already supported.

Implementation note: 

- Since the base() string lifetime is less than the lifetime of targets, the only way I could find to make it work is to use String instead of &str. I tried to add lifetime annotation for the config parameter, but it was not working out (since it's borrowed in some other place). I am novice in rust. So maybe there is a better way. 

- I also modified split_repo_aur_pkgbuilds to return strings instead of the Targ  struct. Indeed, the **repo** field is never used.

### DIFF
--- a/src/download.rs
+++ b/src/download.rs
@@ -10,7 +10,7 @@ use std::process::{Command, Stdio};
 use std::result::Result as StdResult;
 
 use alpm::Version;
-use alpm_utils::{AsTarg, DbListExt, Targ};
+use alpm_utils::{AsTarg, DbListExt};
 use ansi_term::Style;
 use anyhow::{bail, ensure, Context, Result};
 use aur_depends::AurBase;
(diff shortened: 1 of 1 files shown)


arenevier:
test all pass locally for me. (by running `cargo test --features mock`)

Morganamilo:
Sorry I'm so behind on PRs and issues. I fixed this a while ago with c894e25bd7cb19028642b046bf419dc078dd0399

I think it's the better approach as split_repo_aur_pkgbuilds still returns Targs so the caller can choose which part they care about.

PR #1251, opened 2024-09-24
man page : fix typo

Maherilaza:


Morganamilo:
Not sure what's going on here. Fix is good but there's tesdata and translation changes too.

PR #1260, opened 2024-10-11
[Alpm] Fix around upstream alpm issue for aarch64

trcrsired:
See:
https://github.com/archlinux/alpm.rs/issues/35


Morganamilo:
The reason this isn't the default is because it makes clang a build dep.

This is fixed in git so will be fixed in the next release.

PR #1275, opened 2024-11-11
Remove unused crate patches

mkrasnitski:
These patches are not used and cargo gives the following warning when compiling:

```
warning: Patch `alpm v3.0.4 (https://github.com/archlinux/alpm.rs?rev=306342#306342ef)` was not used in the crate graph.
Patch `aur-depends v3.0.0 (https://github.com/Morganamilo/aur-depends?rev=30c2c1#30c2c150)` was not used in the crate graph.
Check that the patched package version and available features are compatible
with the dependency requirements. If the patch has a different version from
what is locked in the Cargo.lock file, run `cargo update` to use the new
version. This may also occur with an optional dependency that is not enabled.
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2613,13 +2613,3 @@ dependencies = [
  "quote",
  "syn",
 ]
-
-[[patch.unused]]
-name = "alpm"
-version = "3.0.4"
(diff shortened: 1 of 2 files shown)


Morganamilo:
Just remove the two causing the warning. I frequently use the others to test cross library changes so I keep them around.

PR #1288, opened 2024-12-11
feat(pkgbuild): add PKGBUILD editing functionality

### LINKED ISSUE #1280: Weird output when using in-terminal file manager
### Affected Version

`paru v2.0.4 - libalpm v15.0.0`

### Description

Just a minor thing, but the output seems to be interrupted and overwritten when using a terminal file manager like `nnn` that opens in the same terminal window as paru. It becomes visible when you close the file manager which you reviewed the changes with. One can continue as usual, but the output just looks weird. Also, as you write over the package name, it is not really clear what you are inputting to the `Accept changes?` prompt.

It does not seem to be `nnn`, because it's the same with `ranger`.

IIRC, this was introduced with `v2.0.0`.

### Output

```sh
$ paru --fm nnn -Su

[…]

Aur (2)                    Old Version  New Version    Make Only
aur/discord_arch_electron  0.0.75-1     0.0.76-1       No
aur/hibiscus               2.10.21-1    2.10.23-1      No

:: Proceed to review? [Y/n]:

:: Downloading PKGBUILDs...
:: Accept changes? [Y/n]: on-0.0.76-1                [--------------------------------------------------]
```

When the paru git clones are up to date, it does not happen:

```sh
:: Downloading PKGBUILDs...
 PKGBUILDs up to date
:: Accept changes? [Y/n]:
```

`paru.conf` probably does not matter, but anyway:

```conf
[options]
BatchInstall
SaveChanges

[bin]
GitFlags = -c user.name=paru_cache
Sudo = doas
```

romanstingler:
Add support for editing PKGBUILDs before building packages, similar to yay's --editmenu option. This allows users to review and modify PKGBUILDs before they are built.

The implementation includes:
- New editor selection logic that respects --editor flag, $VISUAL, and $EDITOR
- Command line options --editmenu, --editor, and --editorflags

Usage:
  paru --editmenu -S package            # Enable PKGBUILD editing
  paru --editor vim --editmenu package  # Use specific editor
  
  might 
  closes  #1280


romanstingler:
I started coding in Rust not long ago, so productive feedback is highly appreciated. 

Basically, I used yay, and it has some features that **Paru** doesn't have, or at least not as comfortable. 

With this patch, ANY EDITOR can be used to edit the `PKGBUILD` file.

Morganamilo:
I'm really not a fan of yay style menus. Paru was a move to get away from them :P

You can use the --fm flag to edit pkgbuilds so this feature is already covered in my opinion.

PR #1289, opened 2024-12-12
chroot: add support for XDG_CONFIG_HOME/pacman/makepkg.conf

### LINKED ISSUE #1283: paru not using `$XDG_CONFIG_HOME/pacman/makepkg.conf` inside CHROOT
### Affected Version
paru v2.0.3.rebuild.1.r1.g90656a9 - libalpm v15.0.0

### Description
I've successfully setup building inside the chroot, but paru doesn't use my `$XDG_CONFIG_HOME/pacman/makepkg.conf` for building.

This leads to errors, as I've set `e.g.` `PKGEXT='.pkg.tar'` which is used locally but not for building, which in turn fails with `failed to run: repo-add -R /var/cache/aur-repo/aur.db.tar /var/cache/aur-repo/$package.pkg.tar`

As far as I can tell paru mounts `/etc/pacman.conf` and `/etc/makepkg.conf` but nothing more.

I know about `/etc/makepkg.conf.d/*.conf` but putting my `makepkg.conf` in there doesn't work by itself (I guess it won't be copied?).

Setting the `paru.conf` option `MakepkgConf = /etc/makepkg.conf.d/custom.conf` fails completely because this flag expects a *full* `makepkg.conf` and not just the changes to the default. (I found this out as I don't have `CARCH` set, which is set by default).

Manually copying the config inside the chroot to `/etc/makepkg.conf.d/custom.conf` works a little bit, but ultimately fails because I changed the `BUILDDIR` inside my `makepkg.conf` which is not picked up by whatever runs inside the chroot.

https://github.com/cwrau/linux-config/blob/eefea8d7cc7e4ca4ae0124eed979171d9aac6225/.config/paru/paru.conf

romanstingler:
Allow users to override makepkg.conf settings by placing a config file in $XDG_CONFIG_HOME/pacman/makepkg.conf. When present, this config will be mounted read-only in the chroot at /etc/makepkg.conf.d/xdg.conf, allowing its settings to take precedence over the base config.

Closes #1283

### DIFF
--- a/src/chroot.rs
+++ b/src/chroot.rs
@@ -92,6 +92,19 @@ impl Chroot {
             cmd.arg(format!("{}.d:/etc/makepkg.conf.d", self.makepkg_conf));
         }
 
+        let xdg_config = std::env::var_os("XDG_CONFIG_HOME")
+            .map(|p| PathBuf::from(p).join("pacman/makepkg.conf"));
+
+        if let Some(xdg_config) = xdg_config {
+            if xdg_config.exists() {
+                cmd.arg("--bind-ro");
+                cmd.arg(format!(
+                    "{}:/etc/makepkg.conf.d/xdg.conf",
+                    xdg_config.display()
+                ));
+            }
+        }
+
         for file in &self.ro {
             cmd.arg("--bind-ro");
             cmd.arg(file);


romanstingler:
I tried with this quick and dirty script
```
#!/bin/sh
set -e

# Test directory setup
TEST_ROOT="/tmp/paru_test"
CHROOT_DIR="$TEST_ROOT/chroot"
XDG_CONFIG_HOME="$TEST_ROOT/config"
export XDG_CONFIG_HOME

# Clean up previous test if exists
rm -rf "$TEST_ROOT"

# Create test directories and configs
mkdir -p "$CHROOT_DIR"
mkdir -p "$XDG_CONFIG_HOME/pacman"

# Create test makepkg.conf with distinctive settings
cat > "$XDG_CONFIG_HOME/pacman/makepkg.conf" <<EOF
PKGEXT='.pkg.tar'
BUILDDIR=/tmp/custom_build
PACKAGER="Test User <test@example.com>"
EOF

# Initialize the chroot with basic packages
echo "Creating chroot environment..."
sudo pacstrap -c "$CHROOT_DIR" base

# Create a basic system makepkg.conf
sudo tee "$CHROOT_DIR/etc/makepkg.conf" > /dev/null <<EOF
PKGEXT='.pkg.tar.zst'
BUILDDIR=/tmp/makepkg
PACKAGER="System <system@example.com>"
EOF

# Copy XDG config to chroot
sudo mkdir -p "$CHROOT_DIR/etc/makepkg.conf.d"
sudo cp "$XDG_CONFIG_HOME/pacman/makepkg.conf" "$CHROOT_DIR/etc/makepkg.conf.d/xdg.conf"

# Test config precedence
echo "=== Testing makepkg.conf settings ==="
sudo chroot "$CHROOT_DIR" /bin/bash <<'CHROOT_SCRIPT'
echo "Base makepkg.conf:"
cat /etc/makepkg.conf
echo -e "\nXDG config makepkg.conf:"
cat /etc/makepkg.conf.d/xdg.conf
echo -e "\nTesting variable precedence:"
. /etc/makepkg.conf
if [ -d /etc/makepkg.conf.d ]; then
    for conf in /etc/makepkg.conf.d/*.conf; do
        if [ -f "$conf" ]; then
            echo "Sourcing $conf"
            . "$conf"
        fi
    done
fi
echo -e "\nFinal values:"
echo "PKGEXT=$PKGEXT"
echo "BUILDDIR=$BUILDDIR"
echo "PACKAGER=$PACKAGER"
CHROOT_SCRIPT
```

Morganamilo:
Feature is good but I don't think it's implemented right. AFAIK makepkg.conf in home will take priority and if that exists no .d config files will be read.

So I think we should look for `~/.makepkg.conf` and ` $XDG_CONFIG_HOME/pacman/makepkg.conf` in that order. If we find either, mount it to /etc/makepkg.conf and leave out the real /etc/makepkg.conf and .d dirs.

Also `dirs::config_dir()` should be used instead of the env lookup. dirs should also do homedir too I believe.

Morganamilo:
Oh I just realised the user makepkg.conf extends the system not replaces it. And the .d config files not applying is only when you manually specify --config to makepkg.

So maybe this approach is the best? I'm not sure the priority of the main config file vs .d dropins though.

PR #1292, opened 2024-12-15
Add coloring to repo names and version diff

### LINKED ISSUE #1291: add colors for repo names and version differences
add coloring of repo names  and version diffs like in yay

romanstingler:
closes #1291

I have also the cachyOS repositories installed 
and the colors are too often the same.
Therefore, I changed to use the repo name length, which appears to be different for each repo.

I had to add some code to calculate the length without the ANSI, so the padding would still be working as intended.


src/fmt.rs
```diff
@@ -443,13 +439,38 @@ pub fn print_install_verbose(config: &Config, actions: &Actions, devel: &HashSet
         });
 
         for pkg in &install {
-            println!(
```
Morganamilo, line 446: if println is going to overpad by the amount of characters we added by colouring. Can't we just do `package_len -= pkg_str.len() -  pkg.pkg.db().unwrap().name().len() - pkg.pkg.name().len() - 1` to calculate the difference like you have but keep using the `{:<package_len$}` and similar formatters instead of repeating strings.

Morganamilo (commented):
I'm not 100% sure on this. It's a lot of code (though i think can be shrunk down considerably).

The main issue is paru's goal is to look like pacman and this breaks away from that. It's not too outlandish like the upgrade menu and other yay formatting so I''m just unsure at the moment.

Maybe if for the diffs we used the gray pacman does when printing versions it would fit in more.

romanstingler:
> Maybe if for the diffs we used the gray pacman does when printing versions it would fit in more.
Hi, sorry but I don't get this sentence.

Only for the version number diffs, 
Or not color the version numbers
Or something different I don't get 
Thx

Morganamilo:
image

Pacman uses these colors to seperate the name from version. We could do the same thing for version diffs. Not sure how it will look though.

PR #1293, opened 2024-12-15
fix: handle EOF in user input prompts

### LINKED ISSUE #1190: Exit on Receving EOF
Currently paru continues to next step if it receives EOF.

romanstingler:
- Add proper EOF detection in ask() function
- Print " -> EOF" message when EOF is received
- Treat EOF as negative response
- Handle input read errors explicitly

closes #1190

image
image

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -134,16 +134,27 @@ pub fn ask(config: &Config, question: &str, default: bool) -> bool {
     }
     let stdin = stdin();
     let mut input = String::new();
-    let _ = stdin.read_line(&mut input);
-    let input = input.to_lowercase();
-    let input = input.trim();
+    match stdin.read_line(&mut input) {
+        Ok(0) => {
+            println!(" -> EOF");
+            false
+        }
+        Ok(_) => {
+            let input = input.to_lowercase();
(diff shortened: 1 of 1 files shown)


src/util.rs
```diff
@@ -134,16 +134,27 @@ pub fn ask(config: &Config, question: &str, default: bool) -> bool {
...
-
-    if input == tr!("y") || input == tr!("yes") {
-        true
-    } else if input.trim().is_empty() {
-        default
-    } else {
-        false
+    match stdin.read_line(&mut input) {
+        Ok(0) => {
+            println!(" -> EOF");
```
Morganamilo, line 139: Pacman doesn't print anything for this. In fact it doesn't even print a new line.

Paru tries to match pacman when possible so negative on EOF is good but I don't think we should print anything but a new line. Otherwise it kind of messes up the input. I'll see about changing pacman to be like that too.

romanstingler: I agree that this should be handled in Pacman. 
Thanks for your time reviewing my stuff and giving feedback. 
I will try to get all PRs fixed as soon as possible.

PR #1298, opened 2024-12-16
update scraper crate dependency

romanstingler:


Morganamilo:
I pull dependency updates intermittently or before release. Unless there's a specific thing the update fixes or a code change paru needs to make for the update. These sort of PR's don't really achieve anything.

PR #1301, opened 2024-12-16
add PostBuildCommand

### LINKED ISSUE #1300: [Feature Request] Implement PostBuildCommand
**Have you checked the readme and man page for this feature?**
**Have you checked previous issues for this feature?**
I have this script I'd like to execute in the working directory of a built package that pushes the package to my self-hosted package cache. Could it be possible to implement a `PostBuildCommand` argument to have paru automatically invoke it for me?

romanstingler:
I have implemented the post-build command feature. Here's what I did:

1. Added post-build command parsing in `command_line.rs`:
   - Added handling for the `--post-build-command` argument
   - The argument requires a value which will be stored in `config.post_build_command`

2. Created a `post_build_command` function in `install.rs`:
   - Similar to `pre_build_command`, it executes a shell command after successful package build
   - The command has access to `PKGBASE` and `VERSION` environment variables
   - The command runs in the package build directory

3. Added post-build command execution in `install.rs`:
   - Called `post_build_command` after a successful package build
   - Only executes if a post-build command is configured

4. Added post-build command config in `config.rs`:
   - Added `PostBuildCommand` option to the config file parser
   - The command can be specified in the config file under the `[bin]` section

Users can specify a post-build command in these ways:

1. Via command line: `paru --post-build-command "your-command"`
2. Via config file: 

```ini
[bin]
PostBuildCommand = your-command
```

The command will receive:

- Current working directory set to the package build directory
- `PKGBASE`: The base package name
- `VERSION`: The package version

---

Tested with 
`cargo run -- -S vim-git --post-build-command "/tmp/code/paru/test.sh"`
and 
`paru.conf`
```
#
# Binary OPTIONS
#
[bin]
PostBuildCommand = "/tmp/code/paru/test.sh"
```

closes #1300

### DIFF
--- a/src/command_line.rs
+++ b/src/command_line.rs
@@ -366,6 +366,12 @@ impl Config {
             }
             Arg::Long("nosign") => self.sign = Sign::No,
             Arg::Long("nosigndb") => self.sign_db = Sign::No,
+            Arg::Long("post-build-command") => match value {
+                Ok(v) => {
+                    self.post_build_command = Some(v.to_string());
+                }
+                Err(_) => bail!(tr!("argument --post-build-command requires a value")),
+            },
             Arg::Long(a) if !arg.is_pacman_arg() && !arg.is_pacman_global() => {
                 bail!(tr!("unknown option --{}", a))
             }
@@ -448,6 +454,7 @@ fn takes_value(arg: Arg) -> TakesValue {
         Arg::Long("sign") => TakesValue::Optional,
         Arg::Long("signdb") => TakesValue::Optional,
         Arg::Long("mode") => TakesValue::Required,
+        Arg::Long("post-build-command") => TakesValue::Required,
         _ => TakesValue::No,
     }
 }
--- a/src/install.rs
+++ b/src/install.rs
@@ -607,6 +607,9 @@ impl Installer {
 
         self.add_pkg(config, base, repo, &pkgdests, &debug_paths)?;
         self.queue_install(base, &pkgdests, &debug_paths);
+
+        post_build_command(config, dir, base.package_base(), &version)?;
+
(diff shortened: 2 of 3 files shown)


src/command_line.rs
```diff
@@ -366,6 +366,12 @@ impl Config {
             }
             Arg::Long("nosign") => self.sign = Sign::No,
             Arg::Long("nosigndb") => self.sign_db = Sign::No,
+            Arg::Long("post-build-command") => match value {
```
Morganamilo, line 369: TakesValue::Required handles the error if a value is not provided. You can just do `value?` to get the value.

src/install.rs
```diff
@@ -607,6 +607,9 @@ impl Installer {
 
         self.add_pkg(config, base, repo, &pkgdests, &debug_paths)?;
         self.queue_install(base, &pkgdests, &debug_paths);
+
+        post_build_command(config, dir, base.package_base(), &version)?;
```
Morganamilo, line 611: I think the position is good, only difference to placing this 2 lines up is it will run after adding the package to the local repo. Which some person may find use in and there's no downside.

src/install.rs
```diff
@@ -1558,6 +1560,19 @@ fn pre_build_command(config: &Config, dir: &Path, base: &str, version: &str) ->
     Ok(())
 }
 
+fn post_build_command(config: &Config, dir: &Path, base: &str, version: &str) -> Result<()> {
```
Morganamilo, line 1563: This is basically identical to the function above. Maybe we turn this into just build_command() where we pass in the command then have tiny functions for pre and post build.

chovanecadam:
This is useful feature. I noticed that manpage is missing, see !539 for reference.

Issue #613, opened 2021-12-09
Removing cargo deps


Morganamilo:
There's a couple of deps that could be removed eventually which should help with compile time + binary size:

```toml
htmlescape = "0.3.1"                                             # aur rpc should get comment support eventually
kuchiki = "0.8.1"                                                # aur rpc should get comment support eventually                          
once_cell = "1.8.0"                                              # should be stable in std eventually
rss = { version = "2.0.0", default-features = false }            # defer to a binary to get the news, are there any good ones?
```

Aditionally reqwest/tokio could be dropped with a little bit of headache

```toml
reqwest = { version = "0.11.6", features = ["gzip", "socks"] }   # could use curl instead but would lose async support
tokio = { version = "1.14.0", features = ["process"] }           # would lose async support
futures = "0.3.18"                                               # not needed if no tokio
```

Curl could be used instead of reqwest. This is good as pacman already deps curl. One issue is we may lose async web requests which we currently make use of to do some requests in parallel (mostly devel checks + update checks). There also appears to be a tokio async provider but that doesn't seem too popular so I wouldn't want to reply on it.

Sync curl would let us also drop tokio and the whole async stack but we would have to use a threadpool for parallel requests. Realisitcly we're probably never doing more than 30 requests. Spawning 30 threads probably is not that bad.

leshow:
Stumbled on the project, chiming in as I've worked with tokio a bunch. I don't think this is an application that really benefits from it so removing it is probably a good idea. On some systems you're going to end up spawning close to 30 threads anyway because the runtime is instantiated with:

```
#[tokio::main]
async fn main() {}
```

This will spawn 1 thread per logical CPU for the scheduler & an additional threadpool for synchronous tasks (stuff you'd run in `spawn_blocking`). So you're likely spawning a bunch of threads that aren't used anyway. Overall, unless there's something I'm missing, you're not getting much benefit from tokio and things may even be faster without it.

For alternatives, if you don't want to use curl you can also go with something like [ureq](https://github.com/algesten/ureq) too.

Morganamilo:
Paru doesn't use any spwawn's or spawn blocking. The only real use for async is selecting and joining network reqests.

leshow:
Right, sounds like even more reason to switch then! It would likely enable the removal of `async_trait` too?

Morganamilo:
Not sure if that's more of a reson. Right now it's all multi layerd. The join is ran on a futures that use select internally.

[...]

PR #1304, opened 2024-12-22
Refactor regex matching logic and migrate to regex-lite crate

romanstingler:
- Replaced `regex` crate with `regex-lite` crate
- `regex-lite` lacks `RegexSet` functionality therefore =>
- Adjusted logic to match against all regexes explicitly.
- Introduced `matches_any_regex` closure to encapsulate regex match checks.
- Added `iter_matches_any` helper function to handle iterables.
- Reduced duplication in the logic for checking matches across pkg fields.

image

#613

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -144,18 +144,18 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.93"
+version = "1.0.95"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


Morganamilo:
Is this noticeably slower at all? The RegexSet was the data structure perfectly suited for this and I assume it had some optimisations over just a Vec<Regex>.

Also looking at our dependency tree:
- `globset` pulls in some regex stuff
- `tr` pulls in the regex crate

So this doesn't actually eliminate regex but pulls in another crate along side it.

PR #1321, opened 2025-01-23
Add missing dependency to `debugedit` in install section

robinmoussu:


Morganamilo:
Debugedit is part of base-devel

robinmoussu:
That's weird, I'm quite sure I checked that I had base-devel installed. Thanks for double checking that it was indeed already covered.

PR #1354, opened 2025-04-17
add git as pacman dependency for installation script

ky0n:


parkerlreed:
Doesn't that need to be *after* the `--needed`?...

ky0n:
it appears to not matter at which position the --needed is it can even be after both git and base-devel. 
but to make it clearer it put git after the --needed.

PR #1388, opened 2025-07-18
feat: don't add indents if a pager is configured, or it won't work with e.g. *Delta*

### LINKED ISSUE #1387: Added leading indents causes *Delta* to stop working as a pager
### Affected Version
**paru -V**

`paru v2.1.0 - libalpm v15.0.0`

### Description

In *Paru* v2.1.0, indents are added to the diff at <https://github.com/Morganamilo/paru/blob/v2.1.0/src/install.rs#L1762>. This causes external pagers like *Delta* to fail. *Paru* v2.0.4 didn't have indents.

### Output

Removing the leading indent manually this way in `/etc/paru.conf` works, but this is more of a hack. I think it's best for them to not be added in the first place for external pagers.

```ini
Pager = sh -c 'sed --expression="s/^    //g" | delta'
```

stevenxxiu:
Closes #1387.

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -1713,7 +1713,9 @@ pub fn review(config: &Config, fetch: &aur_fetch::Fetch, pkgs: &[&str]) -> Resul
             let diffs = fetch.diff(&has_diff, config.color.enabled)?;
 
             if printed {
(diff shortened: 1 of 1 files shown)


AmmoniumX:
Thank you for this PR! It fixes my issue trying to set `delta` as the pager.

Issue #1447, opened 2025-11-24
Panic when listing AUR packages in AUR repository


justbispo:
### Affected Version
**paru -V**
`paru v2.1.0 - libalpm v15.0.0`

### Description
**Have you checked previous issues?**
Yes

Today I've started to get an error on `paru -Sl aur`.

It happens on both the latest release and `-git` version.

I wonder if there's a package with a weird name in the AUR. I've tried to check the last modified packages but, at first glance, didn't see any package with weird characters.

### Output

```sh
thread 'main' panicked at /home/bispo/.local/share/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/alpm-4.0.3/src/db.rs:144:39:
called `Result::unwrap()` on an `Err` value: NulError(0, [0, 14, 185, 20, 190, 28, 253, 203, 149, 49, 6, 199, 174, 122, 104, 43, 232, 83, 165, 169, 102, 122, 167, 45, 5, 124, 57, 227, 152, 6, 31, 172, 77, 225, 246, 229, 6])
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
aur ���˕1Ǯzh+�S��fz�-|9��M��� unknown-version
```

Edit: Less than an hour after, the error is still the same but the undecipherable name is different:
```sh
thread 'main' panicked at /home/bispo/.local/share/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/alpm-4.0.3/src/db.rs:144:39:
[...]

PR #1449, opened 2025-11-25
Don't rely on transparent decompression in list_aur

espeer:
The current AUR package list code relies on the reqwest client library to transparently decompress the packages.gz file. This only works if the server provides the 'Content-Encoding: gzip" HTTP header, which appears to have changed recently.

This patch explicitly decompresses the package list if no encoding information is provided by the server.

Fixes: https://github.com/Morganamilo/paru/issues/1447

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1509,6 +1509,7 @@ dependencies = [
  "cini",
  "dirs",
  "env_logger",
+ "flate2",
  "futures",
  "globset",
  "htmlescape",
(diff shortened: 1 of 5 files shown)


q234rty:
There are two other uses of `packages.gz` in `completion.rs` and `search.rs` which both need to be fixed.

espeer:
Pushed v3 to address completion.rs and search.rs.

navoyan:
In case someone manually syncs this patch and wonders why it doesn't fix anything, try deleting `~/.cache/paru/packages.aur`.
Thank you for the fix, BTW

rsharifnasab:
Just a quick follow-up on this PR. It would be very helpful for my use case. Are there any outstanding issues or feedback needed before it can move forward?

PR #1455, opened 2025-12-11
Change default sort mode to BottomUp (AUR-first) and update manpage

hereparvezali:
What: Change default sort mode to BottomUp so interactive searches show AUR packages first. Update man/paru.8 to reflect new default.

Why: Matches common user expectations (yay-like AUR-first). Users can still override via CLI flags or config.

Testing: Built with cargo build --release and verified interactive behavior.

### DIFF
--- a/man/paru.8
+++ b/man/paru.8
@@ -461,7 +461,7 @@ Don't skip the review process.
 .TP
 .B \-\-upgrademenu
 Show a detailed list of updates in a similar format to pacman's VerbosePkgLists
-option. (See 
+option. (See
 .BR pacman.conf(5)).
 Upgrades can be skipped using numbers, number ranges, or repo
 names.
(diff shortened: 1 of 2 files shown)


PaddiM8:
> changing the default means users who expect pacman-like behavior would now need to configure around it instead, which seems like trading one group's convenience for another's without a clear benefit.

That is true, but I would bet that the vast majority of users prefer the yay-like behaviour. If I search for something that yields a lot of results (which is most of the time), obviously I want the most relevant result to be the most visible one. If the most relevant result is at the top of the list, I have to scroll all the way at the top to see it, which is not feasible in a lot of cases. I struggle to see why anyone would prefer that and everyone I have talked to about paru vs. yay has agreed with this. I would like to recommend paru to people because it is great in every other way, but it makes more sense to recommend yay since it does not need to be configured to have good UX. I don't think it's bad for a wrapper to improve the behaviour of the program it wraps around.

pfeifferj:
hey @PaddiM8 I talked to Lu about it a few weeks ago, and we landed on this: [#1493](https://github.com/Morganamilo/paru/pull/1493). 

curious to hear your thoughts on this compromise, where the new default behaviour would be `TopDown` ordering for non-interactive operations, and `BottomUp` ordering for interactive menus where users select packages

PaddiM8:
@pfeifferj That sounds ideal to me! Very nice

Issue #1454, opened 2025-12-04
Support for libalpm v16.0.0


LeGmask:
### Affected Version
**paru 2.1.0 and older**

### Description
Pacman v7.1.0 and libalpm v16.0.0 got released in arch-testing as of today.
The new version of libalpm break paru (as intended), and as of now paru is not compatible with libalpm v16:
- patching with patchelf : `paru: symbol lookup error: paru: undefined symbol:  alpm_option_set_disable_sandbox`
- recompiling : `this version of alpm.rs does not support libalpm v16.0.0 only v15.x.x is supported`

### Output

```
paru: symbol lookup error: paru: undefined symbol: alpm_option_set_disable_sandbox
```

```
error: failed to run custom build command for `alpm v4.0.3`

Caused by:
  process didn't exit successfully: `/home/evann/paru/src/paru-2.1.0/target/release/build/alpm-45b5776683cd9bb8/build-script-build` (exit status: 101)
  --- stderr

  thread 'main' (151165) panicked at /home/evann/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/alpm-4.0.3/build.rs:25:13:
  this version of alpm.rs does not support libalpm v16.0.0 only v15.x.x is supported
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

```sh
❯ pacman --version                                                      

 .--.                  Pacman v7.1.0 - libalpm v16.0.0
/ _.-' .-.  .-.  .-.   Copyright (C) 2006-2025 Pacman Development Team
\  '-. '-'  '-'  '-'   Copyright (C) 2002-2006 Judd Vinet
 '--'
                       This program may be freely redistributed under
                       the terms of the GNU General Public License.
```

loxoron218:
Are there any current workarounds for this issue?

marmitar:
There certainly is a workaround.

<details>
<summary><tt>PKGBUILD</tt></summary>

```bash
# Maintainer: Morgan <morganamilo@archlinux.org>
pkgname=paru
pkgver=2.1.0
pkgrel=1.1
pkgdesc='Feature packed AUR helper'
url='https://github.com/morganamilo/paru'
source=("$pkgname-$pkgver.tar.gz::https://github.com/Morganamilo/paru/archive/v$pkgver.tar.gz")
backup=("etc/paru.conf")
arch=('i686' 'pentium4' 'x86_64' 'arm' 'armv7h' 'armv6h' 'aarch64')
[...]

PR #1458, opened 2025-12-13
Update for new alpm version

ewanm89:
Update for new alpm version, depends on [https://github.com/Morganamilo/aur-depends/pull/19](https://github.com/Morganamilo/aur-depends/pull/19) being in crates.io.

Also deleted and added Cargo.lock to .gitignore as then it can resolve crates dependency tree on initial cargo build.

Solves #1454

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ paru.tar.*
 po/*.mo
 locale/
 testdata/pkg/*
+Cargo.lock


befabry:
removing a lock file is chaotic IMO

ArjixWasTaken:
From what I gathered, this PR is supposed to be a draft PR, but you did not mark it like that, and you dislike it when others treat it as a non-draft?

AmmoniumX:
> > This is just like #1456 . It doesn't work, since aur-depends=5.0.0 isn't a real version
> > As michaelbeaumont said, removing Cargo.lock goes outside the scope of such a PR and there is a good reason why it's included by default https://doc.rust-lang.org/cargo/faq.html#why-have-cargolock-in-version-control
> > All I'm saying is that if you really want to remove Cargo.lock and believe you have a valid argument, then you should do that in a separate PR, not attach it as part of something meant to address something entirely different
> 
> Please read the description!!! There is a link!!! We can't do anything until aur-depends is updated, I have a PR there too. And that is where Cargo.lock is a problem.

I'm sorry for misinterpreting your intentions if what you say is true. That being said, the description really makes it sound like you're intentionally removing the `Cargo.logk` file.

I suggest you mark this PR as a draft to make it clear it's not meant to be merged as-is.

Ra2-IFV:
> Bro I don't have keys setup on this machine for the signing yet

Fine I'll do this for you

pfeifferj:
[229ad2ceffdf3cab5bb05abdbe2b51e983742a53](https://github.com/Morganamilo/paru/commit/229ad2ceffdf3cab5bb05abdbe2b51e983742a53)

PR #1463, opened 2025-12-14
Fix: Update dependencies for Pacman 7 / libalpm 15+ compatibility

Horoaw:
Updates Cargo.toml to patch alpm and alpm-utils dependencies to the revision specified in the PKGBUILD. This resolves compilation errors with libalpm v15+ (Pacman7). Cargo.lock has been updated to reflect these changes.


Ra2-IFV:
This is a workaround, not a solution. aur-depends should release 5.0.0 then update Cargo.toml.

PR #1493, opened 2026-01-08
feat(sort): add --auto sorting and restrict to interactive menus

pfeifferj:
Users have requested --bottomup as the default sorting behavior. However, paru wraps pacman and should mimic its behavior, which uses topdown ordering.

As a compromise, introduce a new default "Auto" sort mode.
- TopDown ordering for non-interactive operations
- BottomUp ordering for interactive menus where users select packages

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -257,9 +257,9 @@ dependencies = [
 
 [[package]]
 name = "bumpalo"
-version = "3.19.0"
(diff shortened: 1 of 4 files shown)


giffoniL:
I was about to suggest this (BottomUp by default), it's so much better for package selection! Happy a solution is on the way (aside from manual configuration). :-) good job

PR #1515, opened 2026-02-18
feat: add ability to override configs per-package

### LINKED ISSUE #901: Per Package configs
This is a more complete idea of #188 and #748

The idea is to have add a `packages` or more accurately a  `bases` field.

The env section already exists so with a bases field we could say:

```
# Applies to all packages
[env]
BUILDDIR = foo

# Only applies to spotify and google-chome
[env]
Bases = spotify google-chome
BUILDDIR = foo
PKGDEST = bar
```

Similarly this could also work in options:

```
# Applies to all packages
[options]
MakepkgConf = foo

# Only applies to spotify and google-chome
[options]
Bases = spotify google-chome
MakepkgConf = bar
```

Obviously only certain fields will be overrideable. It doesn't make sense to override IgnoreDevel, NewsOnUpgrade, UpgradeMenu. In fact most fields will not be overrideable which makes me think the section should be called something other than `options`. Maybe `[override]` or `[base]`.

I think a cleaner aproch would to do something like `[env.spotify]` but that would mean you could only do one package at a time and it would lead to duplication as users are likely to want to set common settings for groups of packages.

shssoichiro:
This feature enables users to specify per-package overrides in `/etc/paru.conf`. These overrides can be specified for individual packages or for a group of package.

A per-package override may be defined as follows:

```
[override.package.firefox]
MakepkgConf = /etc/makepkg-pgo.conf
```

This configuration would utilize the specified makepkg config file for only the firefox package.

A group override may be defined as follows:

```
[override.group.lto-builds]
Packages = mesa vulkan-radeon
MakepkgConf = /etc/makepkg-lto.conf
Overrides = { MAKEFLAGS = "-j$(nproc)", CFLAGS = "-O3 -march=native" }
```

This configuration allows specifying multiple packages that will use the same custom config. Both package and group overrides allow specifying either or both `MakepkgConf`, which uses a custom config file, and `Overrides` which directly applies the specified env vars to the build.

Closes #901

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -7,6 +7,7 @@ use crate::pkgbuild::PkgbuildRepos;
 use crate::util::{get_provider, reopen_stdin};
 use crate::{alpm_debug_enabled, help, printtr, repo};
 
+use std::collections::HashMap;
 use std::env::consts::ARCH;
 use std::env::{remove_var, set_var, var};
 use std::fmt;
@@ -380,6 +381,27 @@ impl ConfigEnum for YesNoAllTree {
     ];
 }
 
+#[derive(Debug, Default, Clone)]
+pub struct PackageOverride {
+    pub makepkg_conf: Option<String>,
+    pub env: Vec<(String, String)>,
+}
+
+#[derive(Debug)]
+enum OverrideParseState {
+    Package {
+        name: String,
+        makepkg_conf: Option<String>,
+        env: Vec<(String, String)>,
+    },
+    Group {
+        name: String,
+        packages: Vec<String>,
+        makepkg_conf: Option<String>,
+        env: Vec<(String, String)>,
+    },
+}
+
 #[derive(SmartDefault, Debug)]
 pub struct Config {
     section: Option<String>,
@@ -529,6 +551,9 @@ pub struct Config {
 
     pub env: Vec<(String, String)>,
 
+    pub overrides: HashMap<String, PackageOverride>,
+    override_parse_state: Option<OverrideParseState>,
+
     //pacman
     pub db_path: Option<String>,
     pub root: Option<String>,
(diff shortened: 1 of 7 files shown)


ninetailedtori:
Does this handle matching for upstream package groups as well? Not sure if we wanted to support that or not, because honestly this works fine, and I'm not too sure they're getting much use anyway (upstream's [groups](https://wiki.archlinux.org/title/Meta_package_and_package_group), I mean!)

shssoichiro:
It currently doesn't. It could be added by having paru resolve the group name in the `paru.conf` file to see if it matches the name of an upstream group, and expand that to the list of member packages in that group, though I also am not sure if that feature would be useful enough to implement. Or if it might cause confusion (someone wants to make a group named "plasma" not knowing it would apply to all packages in the "plasma" group upstream).

ninetailedtori:
Hmm, yeah I reckon you're right, keep it simple and have it user defined! Besides, for those who do want the packages in a group, they can find it rather easily on the arch repo sites :]

src/config.rs
```diff
@@ -1158,6 +1214,112 @@ then initialise it with:
...
+                        ));
+                    }
+                    OverrideParseState::Group { packages, .. } => {
+                        packages.extend(value.split_whitespace().map(|s| s.to_string()));
+                    }
+                }
+            }
+            "Overrides" => {
+                let value = value.context(tr!("value can not be empty for key '{}'", key))?;
+                let parsed = parse_overrides_map(value)?;
```
pfeifferj, line 1334: `parse_overrides_map` does `inner.split(',')` before processing key=value pairs, so values containing commas break even when quoted.

i checked out the branch and verified this:

```
// input:  { LDFLAGS = "-Wl,--as-needed" }
// split produces: [`LDFLAGS = "-Wl`, `--as-needed"`]
// second element has no `=`, parse errors out

parse_overrides_map(r#"{ LDFLAGS = "-Wl,--as-needed" }"#)
// => Err("invalid override entry, expected KEY = value")

parse_overrides_map(r#"{ LDFLAGS = "-Wl,--as-needed", CFLAGS = "-O2" }"#)
// => same error
```

`-Wl,` flags in LDFLAGS are very common in the wild. either the parser needs to skip commas inside `"..."` pairs, or the manpage should explicitly document that values cannot contain commas.

shssoichiro: Good catch, this is unintended. I'll add a fix+test for it.

ninetailedtori:
Do we support listing patch files, yet? That might be something we could consider. Having an automated patch function is quite easy, and allowing users to list patchfiles could help a lot when managing packages that might take a while before getting an upstream patch actually pushed. I'd recommend supporting patches JUST for patching PKGBUILD, so it's somewhat freedom for the user to patch the PKGBUILD to then load whatever patches they want for subprojects etc for each repo they want to do it for manually.

EDIT: I'm gonna push a PR to this PR, if that's ok? It'll include the ability to patch the PKGBUILD, using either remote or local patchfiles.

ninetailedtori:
https://github.com/shssoichiro/paru/pull/1 welcoming testing on this one! :3

Issue #1376, opened 2025-06-29
paru -B --noconfirm complains about conflicts for the package it's not installing 


gyscos:
### Affected Version
paru v2.0.4 - libalpm v15.0.0

### Description
When attempting to build (but not install) a package with `paru -G package_name && paru -B --noconfirm package_name`, if the given package would conflict with some installed packages, this error is output:

```
:: Conflicting packages will have to be confirmed manually
error: can not install conflicting packages with --noconfirm
```

If `--noconfirm` is removed, then a single Y/n confirmation is required, but nothing regarding any conflict. This seems to indicate that there's no actual decision to make regarding the conflict, and the error is a false positive. 

### Output

```sh
paru -G paru-bin && paru -B --noconfirm paru-bin
```
Output:
```
:: Querying AUR...

:: Downloading PKGBUILDs...
 (1/1) paru-bin-2.0.4-3                              [---------------------------------------]
:: Generating .SRCINFO for ./paru-bin...
:: Resolving dependencies...
:: Calculating conflicts...
:: Calculating inner conflicts...

:: Conflicts found:
    paru-bin: paru-git (paru)  paru-git (paru)

:: Conflicting packages will have to be confirmed manually
error: can not install conflicting packages with --noconfirm
```

Then, trying without noconfirm:
```sh
paru -B paru-bin
```

Output:
```
:: Generating .SRCINFO for ./paru-bin...
:: Resolving dependencies...
:: Calculating conflicts...
:: Calculating inner conflicts...

:: Conflicts found:
    paru-bin: paru-git (paru)  paru-git (paru)

:: Conflicting packages will have to be confirmed manually

Pkgbuilds (1)  Old Version  New Version    Make Only
./paru-bin                  2.0.4-3        No

:: Proceed with installation? [Y/n]:
fetching devel info...
==> Making package: paru-bin 2.0.4-3 (Sat 28 Jun 2025 22:00:44 EDT)
==> Retrieving sources...
  -> Downloading paru-v2.0.4-x86_64.tar.zst...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 3474k  100 3474k    0     0  6095k      0 --:--:-- --:--:-- --:--:-- 13.1M
==> Validating source_x86_64 files with sha256sums...
[...]

PR #1536, opened 2026-04-19
Fix -B --noconfirm conflict check

gyscos:
As reported in https://github.com/Morganamilo/paru/issues/1376, trying to build a package that would conflict with any installed packages currently works interactively, but not with `--noconfirm`.

https://github.com/Morganamilo/paru/commit/77992030587b69b581dba345749ec378f00d18c8 was trying to fix the problem, but the issue persists on the latest master.

To reproduce, an easy way is trying to build one of paru/paru-bin/paru-git (whichever is not currently installed on the system, or all 3).

Example if `paru` is already installed, trying to build `paru-bin`:
```
cargo run -- -G paru-bin && cargo run -- -B --noconfirm paru-bin
```

Currently on the latest master, this outputs:

```
:: Generating .SRCINFO for ./paru-bin...
:: Resolving dependencies...
:: Calculating conflicts...
:: Calculating inner conflicts...

:: Conflicts found:
    paru-bin: paru

:: Conflicting packages will have to be confirmed manually
error: can not install conflicting packages with --noconfirm
```

Another reproduction example: the `eternalterminal` package. It's a split package with 3 packages: a client, a server, and a package that includes both (unfortunately not a virtual package that depends on both the client and the server). Trying to build it produces internal conflicts with the current master branch, but works with this PR.

```
cargo run -- -G eternalterminal && cargo run -- -B --noconfirm eternalterminal
```

In this branch, this properly builds the package.

This now:
* Properly accounts for all packages being built, not just the latest one (for split packages where we build multiple packages)
* Apply the filter on inner conflicts as well, which is what was being triggered on no-confirm.

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -1380,26 +1380,54 @@ fn check_actions(
         );
     }
 
+    let build_only_targets = if install_targets {
+        None
+    } else {
+        let mut build_only_targets = HashSet::new();
+
+        for base in &actions.build {
+            match base {
+                Base::Aur(base) => build_only_targets.extend(
+                    base.pkgs
+                        .iter()
+                        .filter(|pkg| pkg.target)
+                        .map(|pkg| pkg.pkg.name.clone()),
+                ),
+                Base::Pkgbuild(base) => build_only_targets.extend(
+                    base.pkgs
+                        .iter()
+                        .filter(|pkg| pkg.target)
+                        .map(|pkg| pkg.pkg.pkgname.clone()),
+                ),
+            }
+        }
+
+        Some(build_only_targets)
+    };
+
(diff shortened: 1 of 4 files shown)


pfeifferj:
`testdata/clone/devel-bin/` is missing its `.SRCINFO`. `Mock::new()` walks every subdir of `testdata/clone/` and does `Srcinfo::from_path(dir.path().join(".SRCINFO"))?` on startup, so with just a PKGBUILD it ENOENTs and every test panics before running. that's why ci is red

gyscos:
Aaah I see, was trying to figure out what I broke :)

Will fix that, thanks!

PR #1553, opened 2026-05-18
fix(completion): handle NUL-separated package candidates

Scott-Nx:
Fix Bash completion warning caused by NUL-separated package candidates from `paru -Pc`.

## Problem

When completing package names with Bash completion, for example:

```bash
paru -S gpu-screen-reco<Tab>
````

Bash prints:

```text
bash: warning: command substitution: ignored null byte in input
```

The warning comes from `_paru_pkg()`:

```bash
_arch_compgen "$(paru -Pc)"
```

`paru -Pc` emits NUL bytes. Bash command substitution cannot preserve NUL bytes, so Bash drops them and prints the warning.

On my system:

```bash
$ paru -Pc | LC_ALL=C tr -cd '\000' | wc -c
1734
```

## Fix

Convert NUL bytes to newlines before passing the output into `_arch_compgen`:

```bash
_arch_compgen "$(paru -Pc | tr '\000' '\n')"
```

This preserves candidate separation and avoids the Bash warning.

## Testing

Confirmed that `paru -Pc` emits NUL bytes:

```bash
$ paru -Pc | LC_ALL=C tr -cd '\000' | wc -c
1734
```

Confirmed that converting NUL bytes to newlines removes them:

```bash
$ paru -Pc | tr '\000' '\n' | LC_ALL=C tr -cd '\000' | wc -c
0
```

After applying the patch, Bash completion works without the warning:

```bash
paru -S gpu-screen-reco<Tab>
```

Expected candidates are shown:

```bash
$ paru -S gpu-screen-recorder
gpu-screen-recorder               gpu-screen-recorder-notification  gpu-screen-recorder-ui
```

### DIFF
--- a/completions/bash
+++ b/completions/bash
@@ -41,7 +41,7 @@ _pacman_pkg() {
 
 _paru_pkg() {
   [ -z "$cur" ] && _pacman_pkg Slq && return
-  _arch_compgen "$(paru -Pc)"
+  _arch_compgen "$(paru -Pc | tr '\000' '\n')"
 }
 
 _pacman_repo_list() {


pfeifferj:
I can't reproduce the issue you're trying to fix with this PR

Scott-Nx:
I think this may depend on whether `paru -Pc` emits NUL-separated candidates on the local system.

On my machine, `_paru_pkg()` currently does this:

```bash
_arch_compgen "$(paru -Pc)"
```

and `paru -Pc` contains NUL bytes:

```bash
$ paru -Pc | LC_ALL=C tr -cd '\000' | wc -c
1734
```

Bash command substitution cannot preserve NUL bytes, so Bash drops them and prints:

```text
bash: warning: command substitution: ignored null byte in input
```

Could you check whether `paru -Pc` contains NUL bytes on your system?

```bash
paru -Pc | LC_ALL=C tr -cd '\000' | wc -c
```

If this returns `0`, that would explain why you cannot reproduce the warning locally. If it returns a non-zero value, the warning should be reproducible when the completion path calls `_paru_pkg()`.

This patch is mostly a defensive normalization rather than a behavior change. For systems where `paru -Pc` already produces shell-safe output, completion behavior should stay effectively the same. For systems where `paru -Pc` emits NUL-separated output, converting NUL bytes to newlines before Bash command substitution avoids the warning while keeping candidate separation usable for completion.

The patch changes the input passed to `_arch_compgen` from:

```bash
_arch_compgen "$(paru -Pc)"
```

to:

```bash
_arch_compgen "$(paru -Pc | tr '\000' '\n')"
```

# TL;DR

This is not intended to change completion behavior for systems that already work. It only normalizes NUL-separated output before Bash command substitution, because Bash cannot preserve NUL bytes anyway.

pfeifferj:
my guess is that your `~/.cache/paru/packages.aur` is corrupted with raw gzip bytes

sparshjain265:
I am getting the same warning as @Scott-Nx. `paru -Pc | LC_ALL=C tr -cd '\000' | wc -c` reveals I have 1713 NUL bytes, and this fix gets rid of these warnings for me without any noticeable incorrectness in the behaviour.

Issue #1423, opened 2025-10-06
paru should still proceed to basic pacman operations when it fails to reach AUR


Porkepix:
### Affected Version
**paru -V** `paru v2.1.0.r2.g93d7e76 - libalpm v15.0.0`

### Description
**Have you checked previous issues?** Yes

If AUR servers are affected by some issues and can't be reached such as these days due to the DDoS, `paru` just fails on it and doesn't even offer to proceed over the regular core and extra updates. I know for sure I had pending core/extra updates, but wasn't offered to install them. See logs below.

### Output

```sh
$ paru -Syu
:: Synchronizing package databases...
 core                                               120.4 KiB   521 KiB/s 00:00 [##############################################] 100%
 extra                                                8.0 MiB  11.8 MiB/s 00:01 [##############################################] 100%
:: Looking for PKGBUILD upgrades...
:: Looking for AUR upgrades...
:: Looking for devel upgrades...
error: HTTP status client error (429 Too Many Requests) for url (https://aur.archlinux.org/rpc)
$ paru -Syu
:: Synchronizing package databases...
 core                                               120.4 KiB   484 KiB/s 00:00 [##############################################] 100%
 extra                                                8.0 MiB  13.0 MiB/s 00:01 [##############################################] 100%
:: Looking for PKGBUILD upgrades...
:: Looking for AUR upgrades...
:: Looking for devel upgrades...
error: error sending request for url (https://aur.archlinux.org/rpc): error trying to connect: unexpected EOF
```

Morganamilo:
You can just use pacman or `paru -Syu --repo` for repo upgrades. I'd rather the command be loud and fail obviously in cases like this.

Issue #1055, opened 2023-10-22
Devel updates sometimes fail


gladykov:
### Affected Version
paru v1.11.2.r207.g67cc6f4 - libalpm v13.0.2

### Description
I searched previous issues.

When looking for devel upgrades they randomly time out. What I suspect here:
* github rate limit
* or problem with slow (2 MB/s) connection

Because it depends on number of packages installed from git. It started for me with 4 `-git` packages installed. With 5 packages - 2 timed out.

### Output
```
:: Looking for devel upgrades...
error: timed out looking for devel update: https://github.com/DeaDBeeF-Player/deadbeef.git
```

Morganamilo:
I'd assume it's a mix of a slow connection and the timeout being quite short as to not hang paru for a while.

Issue #697, opened 2022-03-16
429 rate limit when installing kwin-git


Aetf:
### Affected Version
**paru -V**
paru v1.9.3 - libalpm v13.0.1

### Description
I want to try git version of kwin and all related libraries, so I did a fresh install in docker.

It turns out that maybe `kwin-git` has too many AUR dependencies that I'm getting 429 error while building it.

I think it is not IP limiting because I can still install dependencies in smaller steps, e.g. `paru -S kconfig-git`.

Maybe there should be some rate limiting in paru when downloading a large amount of AUR packages?

**Have you checked previous issues?**
I found #520 but that is probably caused by too many calls.

### Output
**Include the FULL output of any relevant commands/configs**
I initially encountered this with `paru -S kwin-git`. But it can also be triggered by smaller packages:
```
[build@d009cc2039a0 tmp]$ paru -S kglobalaccel-git
:: Resolving dependencies...
:: Calculating conflicts...
:: Calculating inner conflicts...

Repo (1) qt5-x11extras-5.15.3+kde+r0-1
Repo Make (5) compiler-rt-13.0.1-1  clang-13.0.1-2  python-pyqt5-sip-12.9.1-1  python-pyqt5-5.15.6-7  sip4-4.19.25-3
Aur (6) kdbusaddons-git-5.83.0_r362.g5cc8df6-1  kconfig-git-5.83.0_r837.g0d6cdae-1  kcoreaddons-git-5.89.0_r1337.ge82b37b6-1  
    kwindowsystem-git-5.83.0_r590.g6b16989-1  kcrash-git-5.83.0_r385.gdf59acb-1  kglobalaccel-git-5.83.0_r444.g9a48818-1

:: Proceed to review? [Y/n]: 

:: Downloading PKGBUILDs...
error: command failed: /home/build/.cache/paru/clone: git clone --no-progress -- https://aur.archlinux.org/kglobalaccel-git:
    Cloning into 'kglobalaccel-git'...
    error: RPC failed; HTTP 429 curl 22 The requested URL returned error: 429
    fatal: error reading section header 'shallow-info'
```

**Don't cut parts of the input always include the FULL thing**

**paru.conf and pacman.conf are usually always relevant**
`paru.conf` and `pacman.conf` are all package defaults.

```sh
#
# $PARU_CONF
# /etc/paru.conf
# ~/.config/paru/paru.conf
#
# See the paru.conf(5) manpage for options

#
# GENERAL OPTIONS
#
[...]

Issue #520, opened 2021-08-30
429 Too Many Requests


rightbraintech:
hi i have done a fresh arch install and i was installing some packages using paru but im getting the above error pls can u give the solution

'error: HTTP status client error (429 Too Many Requests) for url (https://aur.archlinux.org/rpc?arg%5B%5D=Font&arg%5B%5D=Ubuntu-Nerd&arg%5B%5D=Ubuntu-font&arg%5B%5D=i3lock-fancy&arg%5B%5D=picom-git&arg%5B%5D=xidlehook&v=5&type=info)'

Morganamilo:
There's some limit on how much you can sent requests to the aur a day. If you've used the aur a lot you've probably ran into that.

rightbraintech:
but i have done a fresh install

Morganamilo:
Install doesn't matter it's by IP address. Anyway closing unless this happens many times with few paru calls.

jose1711:
for me `paru --gendb` is unable to finish due to this error.

PR #1554, opened 2026-05-19
Add AUR download rate limiting and retry options

matclab:
Add four new config options to mitigate AUR rate limiting and
transient network failures during git clone/fetch:

- AurMaxParallel: cap concurrent AUR downloads (default 20)
- AurDownloadDelay: global throttle between downloads (default 0)
- AurMaxRetries: retries on transient errors (default 3)
- AurRetryDelay: base delay for exponential backoff (default 2)

Each option is available as a CLI flag and in paru.conf, and is
documented in paru.conf(5).

This commit needs associated changes in aur_fetch.rs

### DIFF
--- a/paru.conf
+++ b/paru.conf
@@ -24,6 +24,12 @@ DevelSuffixes = -git -cvs -svn -bzr -darcs -always -hg -fossil
 #UpgradeMenu
 #NewsOnUpgrade
 
+# AUR download rate limiting / retry (see paru.conf(5))
+#AurMaxParallel = 20
+#AurDownloadDelay = 0
+#AurMaxRetries = 3
+#AurRetryDelay = 2
+
 #LocalRepo
 #Chroot
 #Sign
--- a/man/paru.conf.5
+++ b/man/paru.conf.5
@@ -223,6 +223,29 @@ cache to be refreshed every time, while setting this to -1 will cause the cache
 to never be refreshed. Defaults to 7.
 
 .TP
+.B AurMaxParallel = N
+Maximum number of AUR packages to download (git clone/fetch) concurrently.
+Lowering this is the main way to avoid AUR rate limiting. Defaults to 20.
+
+.TP
+.B AurDownloadDelay = N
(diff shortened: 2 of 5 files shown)


pfeifferj:
imo we need to wrap up the discussion on whether this is desirable before blindly implementing *something*

paru.conf
```diff
@@ -24,6 +24,12 @@ DevelSuffixes = -git -cvs -svn -bzr -darcs -always -hg -fossil
 #UpgradeMenu
 #NewsOnUpgrade
 
+# AUR download rate limiting / retry (see paru.conf(5))
+#AurMaxParallel = 5
+#AurDownloadDelay = 1
+#AurMaxRetries = 3
+#AurRetryDelay = 2
+
```
pfeifferj, line 32: these defaults don't match the defaults you documented in the man page

pfeifferj (changes requested):

matclab:
> imo we need to wrap up the discussion on whether this is desirable before blindly implementing _something_

Of course, you're right, and I realized I didn't give much context.

There is some sort of rate limiting on AUR and when doing ` paru --aur -Syu`  , paru hammers the AUR git repo and some download fail, which stop the process.
Hence, the need to be more gentle towards AUR infra.

pfeifferj:
pls check #520, #1423, #1055 and #697

matclab:
> pls check #520, #1423, #1055 and #697

Thank you for the pointers.
I have indeed seen  them, but they are all closed and if I make no mistake, none of them offer a solution.
I was still unable to update all my (numerous) aur packages, because it always fails with a 429 error.

pfeifferj:
my point was that this was shot down three times already

also fwiw your 429s are coming from the rpc info call, not git clones. so this PR throttles the wrong code path in any case

matclab:
I have the feeling I'm a disturbance, and I'm sorry for that.
My intention was to help others by sharing what makes paru working again on my computer.

Have a nice day :sun_with_face:

PR #1566, opened 2026-06-14
feat: add minimum release age gate for AUR packages

### LINKED ISSUE #1563: Delay Update Feature
**Have you checked the readme and man page for this feature?**
Yes

**Have you checked previous issues for this feature?**
Yes

**Feature Request**
Supply chain attacks such as this one are becoming more and more prevalent:  https://lists.archlinux.org/archives/list/aur-general@lists.archlinux.org/thread/FGXPCB3ZVCJIV7FX323SBAX2JHYB7ZS4/

To help users deal with this, I propose a feature similar to what several companies and code repositories are doing which is to take advantage of the fact that most of these supply chain vulnerabilities are discovered and cleaned-up relatively quickly: 
https://socket.dev/blog/npm-introduces-minimumreleaseage-and-bulk-oidc-configuration 

I would love a feature where a user could add a package delay measured in hours to their paru config, and paru would then cease to list or allow installing of packages versions or new packages that are newer than the delay value.  This would need to also include dependencies.  Perhaps there should be a force option/flag to temporarily bypass this without having to tweak the setting.  Then extra security conscious users could set like a week delay on packages and hopefully avoid some of the most obvious package hijack attacks.

pfeifferj:
Supply-chain hijacks of AUR packages are usually caught and reverted within a few days, so installing the very latest version carries the most risk. Let security-conscious users refuse AUR packages whose latest version was published more recently than a configurable age.

`MinimumReleaseAge` (with s/m/h/d/w suffix, bare number means days) blocks install and upgrade when any resolved AUR package, dependencies included, is younger than the threshold, and hides such packages from search. MinimumReleaseAgeExclude and `--nominimumreleaseage` provide opt-outs. Devel packages are exempt since they track upstream commits rather than AUR publish time.

Closes: #1563

### DIFF
--- a/man/paru.8
+++ b/man/paru.8
@@ -443,6 +443,23 @@ cache to be refreshed every time, while setting this to -1 will cause the cache
 to never be refreshed. Defaults to 7.
 
 .TP
+.B \-\-minimumreleaseage <duration>
+Refuse to install, upgrade or list AUR packages published more recently than
+this. The duration is an integer with an optional unit suffix (s, m, h, d, w);
+a bare number means days. Applies to dependencies too; devel packages are
+exempt. See
+.B MinimumReleaseAge
+in paru.conf(5).
+
+.TP
+.B \-\-nominimumreleaseage
+Disable the minimum release age check for this run.
+
+.TP
+.B \-\-minimumreleaseageexclude <packages>
+Packages exempt from the minimum release age check. Accepts glob patterns.
+
+.TP
 .B \-\-sortby <votes|popularity|id|baseid|name|base|submitted|modified>
 Sort AUR results by a specific field during search. Defaults to votes.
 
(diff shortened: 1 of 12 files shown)


shellster:
FWIW, I've tested this fork and it seems to work as expected.  Thank you for this change, I hope it gets merged soon.

One minor change I would recommend:
calling `paru --nominimumreleaseage` should allow you to bypass age gating for all upgrades (without specifying a package).  If you disagree than the error telling you that you have packages that are too new and suggesting `--nominimumreleaseage` should probably say `--nominimumreleaseage <package>` to hint that a package name is required.

pfeifferj:
that's a good point. thank you!

shellster:
Any chance this can get reviewed?

ninetailedtori (approved):

